PR #30, opened 2014-09-29
Added Codecov Support

stevepeak:
I've been chatting with @jfairley about working with [codecov](http://codecov.io/), so I figured to add the support so you can see how this package works. You can see the report (from this pr) [here](https://codecov.io/github/stevepeak/org.immutables?ref=be0337bf3a39e902f1b426f4794c8e2b36d7ecd8).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: java
 install: true
 script: "mvn package"
 before_install: 
-  - sudo pip install codecov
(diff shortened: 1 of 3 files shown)


augustotravillio:
Thank you for this pull request. Unfortunately, we were never planning to use this tool or other coverage tool. Coverage is not a subject of neither bragging nor compliance/enforcement for us. If in future we will consider adding such a tool we will definitely consider _codecov_ as an option.

PR #123, opened 2015-06-21
Allow non-conflicting primitive "from" attributes and still generate "from" builder

trask:
Hi, I get lots of these warnings because I use `from` as an attribute name in several @Value.Immutable classes (e.g. from/to in milliseconds range).  I could change the "from" builder name using @Value.Style, but I like the default "from" builder name (even though I'm not using it yet).  I'm wondering if it's ok to safely allow "from" attributes when they are primitives?  Thanks.


trask:
Please ignore me. I can see the benefit of builders not having 2 single-arg methods with the same name. I will just change the style. Thanks.

elucash:
Hi Trask
Good PRs are always welcome! Of course, having same named, but quite different methods on builder is questionable even if "safe". Just to mention, theres `strictBuilder` style, which, when enabled, prevent generation of `from` methods (because strict builder have "only-once" initialization and from methods facilitate "copy-edit" initialization style, not playing well together).

PR #136, opened 2015-07-17
fixing typo and adding simple name to Javadoc links

pushbit:


elucash:
looks good, should look better now, thanks!

PR #153, opened 2015-08-07
add support for making optionals handle null values.

### LINKED ISSUE #152: Switch from Optional.of to Optional.ofNullable
When we have `java.util.Optional<T>` fields, it would be nice that the `withX` methods could handle `null`, and use `java.util.Optional.ofNullable(x)` instead of `java.util.Optional.of(x)`

ldriscoll:
this fixes #152 @augustotravillio 
Let me know if I've missed something terrifying, or if I've done something out of the standard.  I've not used the generators before.

### DIFF
--- a/value-fixture/src/org/immutables/fixture/style/OptionalWithNullable.java
+++ b/value-fixture/src/org/immutables/fixture/style/OptionalWithNullable.java
@@ -1,0 +1,14 @@
+package org.immutables.fixture.style;
+
+import java.util.Optional;
+import java.util.OptionalInt;
+
+import org.immutables.value.Value;
+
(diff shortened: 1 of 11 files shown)


augustotravillio:
Wow, that's great contribution, thank you, Luke! You were exceptionally good getting though our jungles! @elucash can you take a look and merge this, I will, probably, make some additions afterwards?

elucash:
Looks comprehensive, thanks @ldriscoll!
@augustotravillio could you then check where we need to add `@Nullable` to the generated API

aaylward:
:+1: thanks @ldriscoll

Issue #154, opened 2015-08-07
bad code generated if class has underscores in it (e.g., PAYMENT_TYPE)


The-Alchemist:
...the enum and the ``@Value.Immutable`` are in different packages (forcing the code to have an 'import')

e.g.,

## The Enum

```java
package b;

public enum PAYMENT_TYPE {
    PAYPAL,
    CREDIT_CARD,
    BITCOIN;

}

```

## The @Value.Immutable class

```java
package a;

import org.immutables.value.Value;

import b.PAYMENT_TYPE;

@Value.Immutable
public interface SomeValue {
    PAYMENT_TYPE getPaymentType();
}
```

## Generated Code: ImmutableSomeValue

```java
package a;

import b.PAYMENT;
import com.google.common.base.MoreObjects;
import com.google.common.base.Preconditions;
import com.google.common.collect.Lists;
import java.util.Collection;
import javax.annotation.Generated;
//...........snip.................
```

NOTE that the ``import`` is bad:  
``import b.PAYMENT`` instead of ``import b.PAYMENT_TYPE`` 

I think this also happens if it's a regular class and not just an enum.

elucash:
Thank you for the report! It's definitely something with the imports postprocessor.

The-Alchemist:
@elucash : thanks!  if you point me in the right direction, i can submit a pull request.

elucash:
My first guess is that it's improper reading of source import, starting from
[PostprocessingMachine.java#L147](https://github.com/immutables/immutables/blob/master/generator/src/org/immutables/generator/PostprocessingMachine.java#L147)

Overall idea of import processing is that if we can interpret/rewrite things, we do this, if we do not support something - we leave it as is, apparently we are dropping part of an identifer somewhere while we should copy it as is.

augustotravillio:
Not quite there, but near
@elucash can you take a look at #155 ?

The-Alchemist:
Submitted a pull request (https://github.com/immutables/immutables/pull/156)

elucash:
released in 2.0.18, thanks!

Issue #150, opened 2015-08-06
@JsonSerialize/Deserialize without @JsonProperty generation?


mattclamp:
Is this a feature? If not, can I ask that it is? 

I have detected modules that are supposed to use a certain Property Naming Strategy and the auto generation of `@JsonProperty` is overriding my settings. :(

elucash:
This definitely gets in the way. I've tried to came up more reliable handling, though have no confidence that it good solution either. Thank you for reporting this issue. Could you please check latest snapshot if it makes things any better?

```xml
    <dependency>
      <groupId>org.immutables</groupId>
      <artifactId>value</artifactId>
      <version>2.1.0-SNAPSHOT</version>
    </dependency>
```

in

```xml
  <repositories>
    <repository>
      <id>sonatype-nexus-snapshots</id>
      <name>Sonatype Nexus Snapshots</name>
      <url>https://oss.sonatype.org/content/repositories/snapshots</url>
      <releases>
        <enabled>false</enabled>
      </releases>
      <snapshots>
        <enabled>true</enabled>
      </snapshots>
    </repository>
  </repositories>
```

mattclamp:
It seems to now not add those generated tags, but my fields are losing other annotations (not auto-generated) I had on them, which are also important.

elucash:
Just have couple of questions:

1. Could you please try to use `@Value.Style(passAnnotations={MyAnnotation.class,..}..)` on value type (see [how to apply styles](immutables.github.io/style.html))
2. What kind of target element type is for missing annotation. Do they allow placement on fields or only methods?

This might help resolve the issue, we can make further fixes if we came up with best recipe for the annotation handling in such cases.

ldriscoll:
is there any way that we can have the `passAnnotations` apply to the newly created `Json` object, since the Json object is used for the JsonCreator, and we have some annotations that need to apply to that?

elucash:
linking #162 here

elucash:
Closing as minimal missing functionality was provided in PRs #158 and #162. Released in 2.0.21. Thank you!

PR #156, opened 2015-08-07
fix for classes with underscores

The-Alchemist:
Fix for https://github.com/immutables/immutables/issues/154


The-Alchemist:
Also includes a test that fails for classes that _start_ with an underscore.  That's why the build is failing.

The-Alchemist:
Also includes a test that fails for classes that _start_ with an underscore.  Wasn't sure how to fix that one at the moment.

elucash:
Thank you! I'll merge that! For the classes starting with the underscore we will, probably, leave them fully qualified in the generated source code. We try to read imports from the source only as last resort if we cannot resolve types.

The-Alchemist:
Thanks for merging!  Any ideas on when the next release that includes this fix will be?

elucash:
We will release beta to central after the weekend, and 2.1 release is expected then in a week or two. In a meantime you can try snapshots available from sonatype oss repo (same as described in #150). Thanks!

PR #158, opened 2015-08-17
add ability to pass annotations down to the json layer.

ldriscoll:
this is for the addition to #150 so we can pass extra annotations down to the Json layer.  The reason we want this is because we use [Rosetta](https://github.com/HubSpot/Rosetta) and need to pass these annotations for the rosetta mapper.

Issue #120, opened 2015-06-16
Organize/optimize constructors


augustotravillio:
In lot of cases up to 3 constructors generated

+ builder-constructor
+ parameter-constructor
+ attribute-copy

I don't think we should over-optimize or push everything through a single constructor, but we could sometimes use less constructors where they are essentially equivalent.

elucash:
i don't know if we need such level of sophistication, but if done properly that would give cleaner code for small classes

PR #159, opened 2015-08-19
Generate private default noargs constructor for all immutables

alienisty:
Supports frameworks like MyBatis, Hibernate, etc. and should not cause any problems with other use cases.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -1201,6 +1201,20 @@ return [type.factoryOf]([output.linesShortable][for v in type.settableAttributes
 [/if]
 [/output.trim][/template]
 
+[template generateConstructorNoAttributes Type type Attribute... attributes]
+[for v in attributes, n = v.name]
+[if v.primitive]
+  [if v.boolean]
+this.[n] = false;
+  [else]
+this.[n] = 0;
+  [/if]  
+[else]
+this.[n] = null;
+[/if]
+[/for]
+[/template]
+
 [template generateConstructorDefaultAttributes Type type Attribute... attributes]
 [for v in attributes if not (v.generateDefault or v.generateDerived), n = v.name]
 this.[n] = [emptyImmutableInstanceInferred type v];
@@ -1230,13 +1244,15 @@ this.[n] = [if type.generateJdkOnly][objectsUtilRef type]requireNonNull[else][gu
   private final int ordinal;
   private final Domain domain;
 [/if]
-[if type.useSingleton]
 
   private [type.typeImmutable.simple]() {[output.collapsible]
+[if type.useSingleton]
     [generateConstructorDefaultAttributes type type.implementedAttributes]
+[else]
+    [generateConstructorNoAttributes type type.implementedAttributes]
+[/if]
     [generateAfterConstruction type false]
   [/output.collapsible]}
-[/if]
 [if type.useConstructor and (not type.generateConstructorUseCopyConstructor)]
 
   private [type.typeImmutable.simple]([output.linesShortable][for v in type.constructorArguments][if not for.first],[/if]


elucash:
Hi Alessandro
Thank you for the PR! I've seen you've taken in account reference types, primitives, booleans etc, that's great!
At this point I feel that this is not quite the functionality that we need for immutable types as it adds auxiliary constructor that would be used only by a frameworks, moreover the way that frameworks are going to use it is quite "mutable" one, i.e. create object and then assign field values. While Immutables is far from ideal, we came some extra miles (see #120) to generate minimum stuff that people need and naturally feel quite reluctant to add auxiliary constructors. However, there are also some good news, we are half-way in developing Modifiable/Mutable implementation variation of abstract value type. It will naturally (i.e. by design) have private or public no-arg constructor which will allow it to be used with the frameworks you mentions AFAIK. See issue #130 and the development happens on [modifiables branch](https://github.com/immutables/immutables/tree/modifiables), targeting soon-to-be-released version 2.1. One of the important features will be easy conversion to immutable variant and back to mutable, so this should not complicate usage as we expect.

Please, let us know if it will work for you, will the needs be covered by the promised new functionality?

alienisty:
Hi Eugene,
The frameworks I mentioned can actually initialise final fields with some reflection trickery, but they need a default constructor, with any visibility, to be able to do that. I've seen the mutable generator in the code, but having two representations per objects is not ideal for us. We'd rather have mutable struts with no logic, just public fields, to be used as 'transfer' objects and then turn them into immutables if necessary.
Would you be open to adding an attribute to the annotation in the liking of 'singleton' so that the default constructor could be created on request? Alternatively, a system property could be used to enable the creation of the default constructor.

elucash:
Thank you for sharing your concerns. We can go route of inclusion of a new feature flag as a style.
You can augment PR with the following:
* Add new `Value.Style` attribute such as `boolean privateNoargConstructor() default false`. You will have to jump over some additional hoops to propagate this to the template. You can take PR #158 as an example.
* Make sure that new constructor is generated only when the style is enabled are no other noarg constructor generated by other means such as a singleton etc.

I hope this can work this way, feel free to ask any further questions.

alienisty:
Eugene,
What is your preferred way of signaling incompatible style configurations?
For example, when I enable the style to generate the private default constructor, I don't want to allow the pre-computed hash style, because is probably not going to work properly. Should I output an error from the template? Or is there any other preferred validation method?

elucash:
Sure, I think following strategy would work more or less in line how other similar things handled (while a bit random though ) )
  
*  change `ValueType#isUsePrehashed()` to use new style accessor as additional factor

```java
public boolean isUsePrehashed() {
  return immutableFeatures.prehash()
      && !isGeneratePrivateNoargConstructor();
}
// accessor for template, you can use the name you feel suitable
public boolean isGeneratePrivateNoargConstructor() {
  return constitution.protoclass().styles().style().privateNoargConstructor(); // new style
}
```

* Issue warning about prehash being ignored in the presence of `privateNoargConstructor` style

```java
public final class ValueTypeComposer {
...
  ValueType compose(Protoclass protoclass) {
  ...
    checkAttributeNamesIllegalCharacters(type);
    checkAttributeNamesForDuplicates(type, protoclass);
    checkConstructability(type);
    chechStyleConflicts(protoclass); // <-- new method to check for such conflicts
    return type;
  }

  private void chechStyleConflicts(Protoclass protoclass) {
    if (protoclass.features().prehash()
        && protoclass.styles().style().privateNoargConstructor()) {
      protoclass.report()
          .annotationNamed(ImmutableMirror.simpleName())
          .warning("'prehash' feature is automatically disabled to when 'privateNoargConstructor' style is turned on");
  	}
	}
...
...
}
``` 

I guess, issuing a warning is an appropriate thing to do here as prehashing is an implementation detail and don't influence correctness or consistency of a value object.
Hope it helps

alienisty:
Hi Eugene,
Thank you, I had the rest worked out, just wasn't sure on your preference
on how to handle conflicts.
I'll tweak my changes to implement you suggestions and create a new pull
request.

Regards,
Alex

On Mon, 24 Aug 2015 at 16:09 Eugene <notifications@github.com> wrote:

> Sure, I think following strategy would work more or less in line how other
> similar things handled (while a bit random though ) )
>
>    - change ValueType#isUsePrehashed() to use new style accessor as
>    additional factor
>
> public boolean isUsePrehashed() {
>   return immutableFeatures.prehash()
>       && !isGeneratePrivateNoargConstructor();
> }// accessor for template, you can use the name you feel suitablepublic boolean isGeneratePrivateNoargConstructor() {
>   return constitution.protoclass().styles().style().privateNoargConstructor(); // new style
> }
>
>
>    - Issue warning about prehash being ignored in the presence of
>    privateNoargConstructor style
>
> public final class ValueTypeComposer {...
>   ValueType compose(Protoclass protoclass) {
>   ...
>     checkAttributeNamesIllegalCharacters(type);
>     checkAttributeNamesForDuplicates(type, protoclass);
>     checkConstructability(type);
>     chechStyleConflicts(protoclass); // <-- new method to check for such conflicts
>     return type;
>   }
>
>   private void chechStyleConflicts(Protoclass protoclass) {
>     if (protoclass.features().prehash()
>         && protoclass.styles().style().privateNoargConstructor()) {
>       protoclass.report()
>           .annotationNamed(ImmutableMirror.simpleName())
>           .warning("'prehash' feature is automatically disabled to when 'privateNoargConstructor' style is turned on");
>     }
>     }......
> }
>
> I guess, issuing a warning is an appropriate thing to do here as
> prehashing is an implementation detail and don't influence correctness or
> consistency of a value object.
> Hope it helps
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/immutables/immutables/pull/159#issuecomment-134085134>
> .
>
-- 
Alessandro (Alex) Nistico
alienisty@gmail.com

"Any system like Eternity, which allows men to choose their own future,
will end
 by choosing safety and mediocrity, and in such a Reality the stars are out
of reach."

The End of Eternity - Isaac Asimov

alienisty:
I've created a new pull request with the other preferred approach.
Closing this one.

PR #176, opened 2015-09-16
Add choice for exceptions

ldriscoll:
Added support for throwing a default `? extends RuntimeException` on build and deserialization; also adding support for passing a `Function<String, ? extends RuntimeException>` so that the user can call build with a specific exception to throw when they want something other than the default.

### DIFF
--- a/generator/src/org/immutables/generator/SourceExtraction.java
+++ b/generator/src/org/immutables/generator/SourceExtraction.java
@@ -1,11 +1,5 @@
 package org.immutables.generator;
 
-import com.google.common.base.MoreObjects;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
-import com.sun.tools.javac.code.Symbol.ClassSymbol;
 import java.io.IOException;
 import java.nio.CharBuffer;
 import java.util.Arrays;
@@ -13,6 +7,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+
 import javax.annotation.Nullable;
 import javax.annotation.processing.ProcessingEnvironment;
 import javax.lang.model.element.ExecutableElement;
@@ -20,6 +15,7 @@ import javax.lang.model.element.TypeElement;
 import javax.tools.Diagnostic;
 import javax.tools.FileObject;
 import javax.tools.StandardLocation;
+
 import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.compiler.apt.model.ElementImpl;
 import org.eclipse.jdt.internal.compiler.apt.model.ExecutableElementImpl;
@@ -32,6 +28,13 @@ import org.eclipse.jdt.internal.compiler.lookup.MethodBinding;
 import org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding;
 import org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding;
 import org.eclipse.jdt.internal.compiler.lookup.TypeBinding;
+
+import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.sun.tools.javac.code.Symbol.ClassSymbol;
 
 public final class SourceExtraction {
   private SourceExtraction() {}
--- a/generator/src/org/immutables/generator/SourceExtraction.java
+++ b/generator/src/org/immutables/generator/SourceExtraction.java
@@ -1,5 +1,11 @@
 package org.immutables.generator;
 
+import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.sun.tools.javac.code.Symbol.ClassSymbol;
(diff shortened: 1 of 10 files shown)


mjball:
What's the string argument that gets passed to the `Function<String, ? extends RuntimeException>`?

ldriscoll:
it's the same string that went into the IllegalStateException, something like `Cannot build X, some of required attributes are not set [requiredFieldA]`

elucash:
Hi Luke
Thank you for this and other your contributions!
I'm overall ok with the changes, especially as other my ideas was too conceptually heavy compared to this PR, so I like this better.

Things that I think we can address are mostly about API of the build method with exception factory:

1. Additional build method should have name like `buildOrThrow` or `buildOrFail`. I like short methods and use overloading quite often, just think that in this particular case it will do better with the name other than just 'build' to justify parameter such as exception factory function.
2. Better to put additional build method also under feature flag/style as well. On java7 without guava there's no function type available, pure auto-detection might not be best solution here. If made as a style it will not affect fraction of people who won't use it, but tend to complain about API redundancy.
3.  (Optionally, only if it end up not too complicated) Having new build methods accept null is less than optimal, this can be changed to having private null accepting method of even duplicate checkRequiredAttributes, as original one (before this PR) was JIT optimized (in a way similar to [com/google/common/base/Preconditions](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Preconditions.java#L266)), have no idea how additional null check could influence this or if it's still valid.
4. (Optionally, only if it end up not too complicated) `buildOrThrow` could support checked exceptions as well
```java
<X extends Exception> ImmutableValue buildOrThrow(Function<String, X> exceptionFactory) throws X
```

ldriscoll:
@elucash I think I've mostly covered your concerns
1. I initially chose to do the `build` overloading because the regular `build` does actually throw as well, however I've overcome my concern of that, changed it to a `buildOrThrow` and placed an `@throws` in the docs for the regular one.
2. I've added both the `buildOrThrow` and the `checkRequiredAttributesOrThrow` as additional methods, these are wrapped in requiring java8 or guava, so they should be all gone for java7 no guava people
3. By doing additional methods, they now require non null, would you prefer that to be annotated?
4. I've done the exception check, as suggested but with `T` as it's statically used elsewhere, so there won't be an override issue. 

Additionally the default exception can extend `Exception` if it's a runtime exception then everything is good, if it's not the user must catch it.

ldriscoll:
thanks @elucash let me know if there are other tweaks that should be addressed.

elucash:
Thank you! :+1:
@augustotravillio can you also take a look before we release it?

augustotravillio:
Hey guys, I mostly ok with the change. But would you mind if I add the code to disable the generation of the new build method when `strictBuilder` mode enabled?
We have a lot of functionality for convenience, but for me `buildOrThrow` is slightly redundant and could be replaced with library method like

```java
ImmutableObject.Builder builder = ImmutableObject.builder();
...
ImmutableObject o = getOrThrow(builder::build, IOException::new);

...
static <T, X extends Exception> T  getOrThrow(Supplier<T> supplier, Function<String, X> toException) throws X {
  try {
    return supplier.get();
  } catch (IllegalStateException cannotBuildException) {
    throw toException.apply(cannotBuildException.getMessage());
  }
}
```
Strict builder is on a diet, it's reasonable to not generate new method in a strict mode.

ldriscoll:
I don't mind it if it's disabled under strictBuilder.  The main concern I have with creating a library method is that we'd be creating two traces unnecessarily; secondary to that, having the `buildOrThrow` method makes it clear that the regular `build` throws something.

mjball:
Part of the reason that@ldriscoll and I are interested in this `buildOrThrow` behavior is so that we can end up with a different exception when deserializing JSON from an HTTP request body. In our stack we're using Jackson and Jersey; Jackson propagates the exception thrown by `build()` up to Jersey.

If a more-specific exception than `IllegalStateException` is thrown then we can add, say, a Jersey exception mapper. But the point is, it's library code, not our code, that's calling `build()`, so I'm not sure how we'd incorporate @augustotravillio's `getOrThrow(Supplier, Function)`.

ldriscoll:
@mjball that case is actually taken care of by the style attribute `throwForInvalidImmutableState` the `buildOrThrow` is there so we can throw any chosen exception when we call `build` in general code use

mjball:
So it is, my mistake!

ldriscoll:
The reason I wanted to throw different types of exceptions is because when we deserialize we want a client error, but when we build one in code we'd want a server error

ldriscoll:
@augustotravillio there is another way around the `buildOrThrow`, we could create an interface like this:

```java
public interface ImmutableBuilder<T> {
  T build();
  default T buildOrThrow() {
    try {
      return build();
    } catch (Exception e) {
      throw new RuntimeException(e.getMessage());
    }
  }
}
```

Then in our immutable interface we can do
```java
public interface Point {
  double getX();
  double getY();

  interface Builder extends ImmutableBuilder<ImmutablePoint>() {}
}
```

This seems to work, though I would much prefer that our `interface Point` could extend an interface that contains the whole Builder interface.  If you really don't want the extra builder code in there, we could get around it like this.

elucash:
That is actually a great way to solve it, adding a way to detect builder interfaces is not difficult, yet very useful! As for now now we could release `buildOrThrow` in experimental _beta_ mode to try it out. Another area I am looking into is how we could improve situation on an another side, the side where we deserialize value objects. While we outsource most of the de-/serialization to Jackson, we still have some control over deserialization in `@JsonCreator` method. Do you have any ideas how we could accommodate the original goal at that side?

augustotravillio:
Idea with the Builder is great indeed, thank you @ldriscoll for outlining it! I'm ok with everything else (with a strict builder still on a diet :-) )

PR #197, opened 2015-11-10
Fix javadoc throws syntax

trask:
I'm getting the below javadoc error in my Immutable generated classes.

```
error: reference not found
* @throws exception {@code java.lang.IllegalStateException} if any required attributes are missing
^
```


elucash:
Thank you!

Issue #185, opened 2015-10-27
Support Jackson JsonAnyGetter


DavidLeoni:
Seems like it's not possible to store unknown fields in immutable objects when deserializing with jackson. The following test serializes correctly to {"x":"y"}, but then on deserializing throws this exception:

`com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "x" (class trials.ImmutableObj$Json), not marked as ignorable (one known property: "other"])`

```

@Value.Immutable
@JsonSerialize(as = ImmutableObj.class)
@JsonDeserialize(as = ImmutableObj.class)
abstract class Obj {    
    
    @JsonAnyGetter
    @Value.Parameter
    public abstract Map<String, String> other();
    
}


public class MyTest {
    
    
    @Test
    public void test1() throws Exception{
	ObjectMapper om = new ObjectMapper();	
	
	HashMap<String, Object> map = new HashMap();
	map.put("x",  "y");
	ImmutableObj imm = ImmutableObj.builder().putOther("x", "y").build();
	
	String json = om.writeValueAsString(imm);
	
	System.out.println(json);
	
	ImmutableObj ap = om.readValue(json, ImmutableObj.class);
    }
    
}
```

Maybe the generated ImmutableObj.Json class should contain a `@JsonAnySetter` method.

elucash:
It should contain it. Need to add support for this. Thank you for the report!

elucash:
This should be supported now, released in 2.1.1. Thank you!

PR #204, opened 2015-11-20
Add support for annotation groups

ldriscoll:
Rosetta Annotations, are all annotated with `@RosettaAnnotation`; other libraries do this for good housekeeping.  I have made a change here so that our style can have:
`additionalJsonAnnotations = { RosettaAnnotation.class }`

so that when Rosetta adds more annotations we don't need to add them to the style. 

Hope it makes sense

### DIFF
--- a/value-processor/src/org/immutables/value/processor/meta/Annotations.java
+++ b/value-processor/src/org/immutables/value/processor/meta/Annotations.java
@@ -17,6 +17,7 @@ package org.immutables.value.processor.meta;
 
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Target;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
@@ -48,10 +49,11 @@ final class Annotations {
       ElementType elementType) {
     List<CharSequence> lines = Lists.newArrayList();
 
+    Set<String> seenAnnotations = new HashSet<>();
     for (AnnotationMirror annotation : element.getAnnotationMirrors()) {
       TypeElement annotationElement = (TypeElement) annotation.getAnnotationType().asElement();
 
-      if (annotationTypeMatches(annotationElement, includeAnnotations, includeJacksonAnnotations)
+      if (annotationTypeMatches(annotationElement, includeAnnotations, includeJacksonAnnotations, seenAnnotations)
           && annotationMatchesTarget(annotationElement, elementType)) {
         lines.add(AnnotationMirrors.toCharSequence(annotation));
       }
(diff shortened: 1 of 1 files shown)


ldriscoll:
@jhaber

jhaber:
I think this a very deep rabbit hole. Having an intermediary class for deserialization means that immutables will never play nice with Jackson

elucash:
@jhaber it is so true. In fact we now use original class for writing json, and synthetic one only for reading. 
And still have #185 and alike. Jackson on it's own have couple of inconsistencies (like handling of `JsonValue` with regards to type mapping etc).
So while I'm reviewing this pull request, we're also open to suggestions how to radically improve the situation @ldriscoll  @jhaber have any idea?

jhaber:
I'm not very familiar with the generated code, is it not feasible to deserialize directly into the immutable type? If an intermediary type is necessary, you could try making it a POJO that extends the interface/abstract class, that way it inherits any annotations and extra methods

jhaber:
I'm not very familiar with the generated code, is it not feasible to deserialize directly into the immutable type? If an intermediary type is necessary, you could try making it a POJO that extends the interface/abstract class, that way it inherits any annotations and extra methods

ldriscoll:
I think there would need to be a json deserializer that constructed one from a builder, and then called .build().  We'd still need for the annotations to be passed down into the immutable object though.

ldriscoll:
I actually have an idea around this, I might give it a go.

ldriscoll:
ok, I have a good working version of this now, but I'm still cleaning it up, it'll mean we can get rid of the additional json annotations and such.  I'm going to spend some time with the family, if you'd like to take a look the code is here: https://github.com/ldriscoll/immutables/tree/json_extend

ldriscoll:
nope, going to have to leave the additional annotations there, because jackson, and friends, don't like not standard getters, which immutables annotates with `@JsonProperty` even when they are not in the parent class.  My feeling is that if main class is like this: https://github.com/immutables/immutables/blob/master/value-fixture/src/org/immutables/fixture/jackson/SampleJacksonMapped.java

Then it's the user's fault that it won't serialise to json, but currently Immutables puts `@JsonProperty` on the immutable's getters, so it serialises.

If you want to break that 'feature' we could get rid of ALL of the json annotation passing.

elucash:
Thank you for the work! I'll merge those things now, but it will undergone some changes. Now I would take some deep thinking on how we could make all this working together (including #206).
Essentially, you've, probably, pushed this in the right direction, we just need all currently supported case working, including newly reported stuff.

PR #206, opened 2015-11-21
Fix up the Json class so that it extends the main interface

ldriscoll:
This means that, for the most part, we don't need to pass json annotations down into the immutable at all.  Sadly that's not quite the case because we support this case: https://github.com/immutables/immutables/blob/master/value-fixture/src/org/immutables/fixture/jackson/SampleJacksonMapped.java

Where there are non standard getter names, that are NOT annotated here, and the generated immutable looks like this:
```java
public final class ImmutableSampleJacksonMapped implements SampleJacksonMapped {
  private final String a;
  private final ImmutableList<Integer> b;
  private final Optional<SampleJacksonMapped.RegularPojo> pojo;
....

  /**
   * @return value of {@code a} attribute
   */
  @JsonProperty
  @Override
  public String a() {
    return a;
  }
```


ldriscoll:
@jhaber

ldriscoll:
@mjball

value-processor/src/org/immutables/value/processor/meta/ValueType.java
```diff
@@ -57,7 +43,20 @@
 import org.immutables.value.processor.meta.Proto.Environment;
 import org.immutables.value.processor.meta.Proto.Protoclass;
 import org.immutables.value.processor.meta.Styles.UsingName.TypeNames;
-import static com.google.common.base.MoreObjects.*;
+
+import com.google.common.base.CaseFormat;
```
ldriscoll: sorry about the import reformatting.

mjball:
Why does the `Json` class exist in the first place?

ldriscoll:
I initially tried using the json deserialize using a builder, sadly it doesn't call the set methods, it directly sets the fields, and therefore the flags for marking that a set was called, was failing.

elucash:
Thank you! while I merged this, I will also do some "heavy iron" on it to make sure we are backward compatible and sound.

PR #213, opened 2015-12-02
Added support for the Atlassian Fugue Option class

mzeijen:
Atlasssian Fugue is a nice little library that is an functional extension of Guava and/or Java 8. Fugue contains an Option class which is an alternative to the Optional type class of Guava and/or Java 8.

This pull request adds the same support for the Fugue Option class as the other Optional classes.

Atlassian just released Fugue 3 which uses a different package (io.atlassian.fugue) then Fugue 2 and older (com.atlassion.fugue). This pull requests add support for both versions.

### DIFF
--- a/cases/pom.xml
+++ b/cases/pom.xml
@@ -52,6 +52,18 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>com.atlassian.fugue</groupId>
+      <artifactId>fugue</artifactId>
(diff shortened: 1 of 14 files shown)


elucash:
Thank you for the contribution! This will see the light in the next release, hopefully in one or two weeks

mzeijen:
Awesome! Thx.

PR #218, opened 2015-12-07
Jsondeserialize not necessary

ldriscoll:
…e or @JsonDeserialize annotations.

We currently have a style class that is meta annotated with `@JsonDeserialize`, since we use jackson for all of our code.  Sadly, this means that all generated code contains `@JsonDeserialize`, which is technically unnecessary, unless the parent class also is annotated with it.

This PR adds the option to generate the static `Json` class without having to add the annotations.

### DIFF
--- a/value-fixture/src/org/immutables/fixture/jackson/JacksonMappedWithNoAnnotations.java
+++ b/value-fixture/src/org/immutables/fixture/jackson/JacksonMappedWithNoAnnotations.java
@@ -1,0 +1,10 @@
+package org.immutables.fixture.jackson;
+
+import org.immutables.value.Value;
+import org.immutables.value.Value.Style.GenerateJacksonMappings;
+
+@Value.Immutable
+@Value.Style(generateJacksonMappings = GenerateJacksonMappings.ALWAYS)
+public interface JacksonMappedWithNoAnnotations {
+  String getSomeString();
+}
--- a/value-fixture/src/org/immutables/fixture/jackson/JacksonMappedWithNoAnnotations.java
+++ b/value-fixture/src/org/immutables/fixture/jackson/JacksonMappedWithNoAnnotations.java
@@ -1,10 +1,9 @@
 package org.immutables.fixture.jackson;
 
 import org.immutables.value.Value;
-import org.immutables.value.Value.Style.GenerateJacksonMappings;
 
 @Value.Immutable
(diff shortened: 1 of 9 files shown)


elucash:
Hi @ldriscoll, thank you for the efforts! For this PR I have to understand what particular problem we are solving

Currently, we are adding deserialize annotation to `Json` class in order to prevent recursiveness of definition when inheriting `@JsonDeserialize(as = Immutable...`. If we don't inherit the annotation (`Json` is being generated by virtue of having meta annotating style class), then we don't need to override it. But even if we do put `@JsonDeserialize` on `Json` it should be harmless, right? Or will `@JsonDeserialize` on `Json` override some settings that are useful and come from the style meta-annotation? Regardless if it's harmless or not to have such override, it is, of course, makes sense to be safe and accurate and just not generate the override if not needed. But then, we could do this without having a configuration, just correct and reasonable built-in behavior. I somehow assume that it would be safe without having additional configuration.
Please correct me if I missed something important.

ldriscoll:
You have it completely correct, the one use case I have for removing it is this:  We use a single style class, in it's own project, that is then used by multiple projects.  As a rule, all of our immutables should be deserialisable using Jackson, and all of our projects already have jackson-annotations as a dependency; however, adding `@JsonDeserialize` to the generated code, when it is not present in the interface or abstract class, forces the project to require jackson-databind, even though they don't need it.

elucash:
sorry for not acting on it right away, just had to reread to actually understand. Seems absolutely reasonable to not introduce unwanted dependency. Could we just make it without style configuration? I'm yet to see a use-case when it has to be configurable.

ldriscoll:
@mjball was talking about that too; we'd need to detect whether json-annotations is in the classpath first though.

mjball:
:100: for not exposing a separate config if we can make it do the right thing in all cases without more config. But you already knew that @ldriscoll :)

ldriscoll:
I'll have a bash at the detection.

ldriscoll:
@elucash I've changed to use the classpath, but it's not working, since the `serial` pom doesn't include it, but it's in the classpath; I'm going to need some advice here.

elucash:
@ldriscoll @mjball can you, please, take a look at master? do we got it right?

ldriscoll:
no, like I said, it isn't working because the serial pom doesn't contain JsonCreator in it's classpath, but the classpath.available says that it does.

Issue #224, opened 2015-12-14
Modifiable.generator equals method checks instance of this instead of another


akune:
Modifiable.generator templates the equals method in the following way: 
```java
@Override
public boolean equals([atNullable]Object another) {
  if (this == another) return true;
  if (!(this instanceof [equalToType])) return false;
  [equalToType] o = ([equalToType]) another;
  [if type.mandatoryAttributes]
  if (![isInitialized type]() || !o.[isInitialized type]()) {
    return false;
  }
  [/if]
  return equalTo(o);
}
```
The guard 
```java
if (!(this instanceof [equalToType])) return false;
```
should actually check the instance of the another object like this
```java
if (!(another instanceof [equalToType])) return false;
```
The current implementation leads to the following test throwing a ClassCastException: 
```java
import org.assertj.core.api.Assertions;
import org.immutables.value.Value;
import org.junit.Test;

public class ModifiableEqualsTest {

	@Value.Modifiable
	public static interface MyValue {
		String getName();
	}

	@Test
	public void modifiableShouldNotEqualObjectOfDifferentType() {
		MyValue value = ModifiableMyValue.create();
		Assertions.assertThat(new Object()).isNotEqualTo(value);
		// Will throw ClassCastException here: 
		Assertions.assertThat(value).isNotEqualTo(new Object());
	}

}
```

elucash:
Released in 2.1.5. Thank you!

PR #225, opened 2015-12-14
Fix 224 pull request

akune:
Pull request with fix for issue #224.


elucash:
Good catch, thank you!

PR #246, opened 2016-01-27
Don't unbox types in setters for optionalAcceptNullable.

ldriscoll:
@elucash currently the setter accepts `@Nullable int value`, and should accept `@Nullable Integer value`


elucash:
Makes sense :thumbsup:
Thank you!

PR #308, opened 2016-04-22
Make initShim fields transient

dbevacqua:
Fix https://github.com/immutables/immutables/issues/309 by making initShim fields transient


elucash:
Thank you!

PR #331, opened 2016-05-19
Correct spelling mistake in warning message

jon-hanson:
I've corrected the use of "loosing" to "losing" in this warning message:

> Generic wildcards are not supported, so the container attribute _loosing_ its special treatment


elucash:
Thank you! merged

PR #388, opened 2016-07-01
add some details as to the failure, so people can fix it.

### LINKED ISSUE #386: Code not generated in some projects
I'm having a really hard time tracking this down, but we have some projects, that have the correct imports, but are no longer generating the immutable classes.
All was working in 2.1.15, but after upgrading to 2.2.8 the files are just not being created.  I've experimented around, and things work with 2.2, but not 2.2.1.
It's worth noting that the project I'm seeing this on is quite large.

ldriscoll:
resolves #386

### DIFF
--- a/value-processor/src/org/immutables/value/processor/meta/TypeStringProvider.java
+++ b/value-processor/src/org/immutables/value/processor/meta/TypeStringProvider.java
@@ -15,14 +15,13 @@
  */
 package org.immutables.value.processor.meta;
 
-import com.google.common.base.Ascii;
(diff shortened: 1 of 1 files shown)


elucash:
Thank you! I'll see maybe it's better to write this to `Reporter`, but it's not urgent

ldriscoll:
that would definitely be nice.  Sorry, I should have tried using that.

ldriscoll:
also, this is not a critical fix for us now, now that I've found the offending code.

Issue #396, opened 2016-07-13
Constructor annotations with @Target(CONSTRUCTOR)


zsolt-donca:
There is a problem with the constructor annotations feature introduced in #317 , namely, it doesn't work with annotations targeting only constructors; it only works with annotations targeting methods as well.

My use case: I am trying to generate immutable entity types to use with [Spring Data](http://projects.spring.io/spring-data/). For Spring Data to be able to work with the classes, they need to have a single constructor that accepts all the properties of the entity (unfortunately, no static factory methods or other builders seem to be supported). I can make _Immutables_ generate such a constructor easily by using `@Value.Style(of = "new", allParameters = true)`. There is a problem, however: there must be a single constructor; in case there are multiple constructors, one needs to be marked with @[PersistenceConstructor](http://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/annotation/PersistenceConstructor.html). _Immutables_ generates additional (private) constructors used by "withProperty" methods and by the type's builder. 

Using the approach described in #317 (using an abstract class to define the immutable type, and to annotate a protected no-arg constructor with the desired annotation) doesn't seem to work with the annotation `@PersistenceConstructor`, and the reason seems to be that it is defined to target only constructors:
```
@Retention(value=RUNTIME)
@Target(value=CONSTRUCTOR)
public @interface PersistenceConstructor
```

The example written in #317 uses [javax.inject.Inject](http://grepcode.com/file/repo1.maven.org/maven2/javax.inject/javax.inject/1/javax/inject/Inject.java?av=f), that targets _methods_, _constructors_ and _fields_.

I think I have a solution for this problem; so expect a pull request in short time.

elucash:
Released in 2.2.11 Thank you!

keithrooney:
@zsolt-donca I'd be interested in seeing an example of this. I'm trying to pass `org.springframework.data.annotation.PersistenceConstructor` to the `passAnnotations`, as seen below, but the annotation doesn't seem to be placed in the generated source code anywhere. 
```

@Value.Style(
    of = "new", 
    allParameters = true,
    passAnnotations = {PersistenceConstructor.class}
)
@Value.Immutable
public interface Message {

    @Id
    @Nullable
    String getId();

    String getSender();

    String getRecipient();

    String getBody();

}
```

PR #397, opened 2016-07-13
#396 Fix for constructor annotations with @Target(CONSTRUCTOR)

zsolt-donca:
Fixed the generation of constructor annotations by using the element type CONSTRUCTOR instead of METHOD.


elucash:
Great! Thank you! Can you, please, add also compilation test to fixture module `/fixture/src/org/immutables/fixture/*` Just a class with constructor annotation which fails to compile without the fix and succeed with the fix?

PR #398, opened 2016-07-14
Added a new case to the 'annotation passing to constructor' test...

zsolt-donca:
… with an annotation targeting only constructors.

Continues #397 with the associated tests.


elucash:
Thank you!

Issue #420, opened 2016-08-16
unset doesn't work


marquiswang:
"unset" is an option in @Value.Style, but doesn't appear to be actually supported.

It would be pretty convenient when dealing with Modifiable classes to undo a set.

elucash:
It was, probably, meant to be used with Modifiables or Mongo repositories. As of now it's seems that its unused. If you have a good idea how it could be implemented, please, try fork Immutables and sketch it in Modifiables.generator template, I believe you should find there any kind of example how to get attributes, names, types etc, feel free to ask any question about it. Thank you!

tyler-reeves:
Just submitted a pull request, #430, aimed at addressing this issue.

elucash:
@tyler-reeves thank you for the excellent contribution!

elucash:
released in 2.3 Thank you!

PR #430, opened 2016-08-18
Address issue #420

tyler-reeves:
Implement unset functions to reset variables to their initial values.


elucash:
Excellent work! Thank you!

PR #457, opened 2016-10-08
Add support for `@javax.annotation.CheckReturnValue`

Stephan202:
The commit message explains the rationale:

```
Since immutable objects are stateless and the invocation of methods on
immutable objects does not have observable side effects, one should generally
use the result of such invocations. A class-level `@CheckReturnValue`
annotation communicates this fact. Tools such as Error Prone can check for
adherence to this contract [1].

An exception to this rule are various methods defined on the nested Builder
class, as those methods return `this`. Those methods may thus be annotated
`@com.google.errorprone.annotations.CanIgnoreReturnValue`.

This change only applies `@CheckReturnValue` if _both_ `@CheckReturnValue` and
`@CanIgnoreReturnValue` are on the classpath, for application of the former
without the latter needlessly forces users of the nested builder into a fluent
usage style.

[1] http://errorprone.info/bugpattern/CheckReturnValue
```

I'm new to Immutables and would love to receive feedback on this PR. Perhaps I didn't follow best practices in applying the code generation DSL, or perhaps `@CheckReturnValue` is also eligible to be applied to other types of generated classes. If so, let me know, and I'll gladly extend the PR. Thanks for taking this feature into consideration.

### DIFF
--- a/value-fixture/pom.xml
+++ b/value-fixture/pom.xml
@@ -164,6 +164,10 @@
           <compilerVersion>1.8</compilerVersion>
           <source>1.8</source>
           <target>1.8</target>
+          <compilerArgs>
+            <!-- The test fixtures invoke builders without using the results. -->
+            <arg>-Xep:CheckReturnValue:WARN</arg>
+          </compilerArgs>
         </configuration>
         <dependencies>
           <dependency>
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -135,6 +135,9 @@ import static [routine].immutableCopyOf;
 [if classpath.available 'javax.annotation.concurrent.Immutable']
 @javax.annotation.concurrent.Immutable
 [/if]
+[if classpath.available 'javax.annotation.CheckReturnValue' and classpath.available 'com.google.errorprone.annotations.CanIgnoreReturnValue']
+@javax.annotation.CheckReturnValue
+[/if]
 [for a in type.passedAnnotations]
 [a]
 [/for]
@@ -1116,6 +1119,7 @@ public interface BuildFinal[type.generics] {
    * Specify a non-default {@link Domain} for ordinal values.
    * @return {@code this} builder for use in a chained invocation
    */
+  [atCanIgnoreReturnValue]
   public final [builderReturnType type] domain(Domain domain) {
     this.domain = [requireNonNull type](domain, "domain");
     return [builderReturnThis type];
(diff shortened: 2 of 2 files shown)


Stephan202:
The reported build failure is unrelated to this PR, and should be resolved once #458 is merged.

Stephan202:
(Rebased the branch, cause I like seeing a passing build ;).)

elucash:
Thank you @Stephan202 for the hard work! I'm looking into this now. Overall I support the idea, just want to make sure we'll get what we actually want ). First question: did you considered just annotating generated accessors/withers with `CheckReturnValue`, in the sense will it solve the issue? Somehow I'm hesitating to put over-restriction on free-form methods that could be defined on abstract value type (however limited to only people using analyzer such as findbugs/errorprone etc). I, probably, missing something. Let me take another look, maybe I'll come up with new questions or ideas

Stephan202:
Thanks for considering this, @elucash. I did briefly consider doing it "the other way around", but went for the current approach for two reasons:

1. In my experience, when writing code with a heavy focus on immutability and functional programming, `@CheckReturnValue` applies to "almost all methods". In my own code I generally apply this annotation at the package level. I rarely use `@CanIgnoreReturnValue` myself, but when I do it's almost always on builder methods.
2. In the Immutable case in particular, it seems the methods for which `@CanIgnoreReturnValue` make sense are limited (namely just the builder methods), while for all other methods `@CheckReturnValue` makes sense:

    1. Factory methods (including those for singletons).
    2. Copy methods.
    3. Builder constructors.
    4. The builder finalizers (`#build()`).
    5. The accessor methods.
    6. The "withers".
    7. `#equals`/`#hashCode`/`#toString`.
    8. Various non-`void` private methods (this doesn't matter from an API perspective, admittedly). 

That said: there may be some Immutables features I haven't considered. What I did try to do is carefully analyze all uses of `this` in `Immutables.generator`. Based on that I _think_ the current application of `@CanIgnoreReturnValue` is exhaustive. I did however not check interaction with the other `*.generator` classes. (Though when I use the augmented code in a private project, Error Prone is happy.)

elucash:
I'm willing to trust your evaluation here, also as there are not a lot of people having `CanIgnoreReturnValue` in the classpath. If there will be legitimate complaints, we'll make refinements to how it's enabled. Thank you!

Stephan202:
Thanks. If I find a better way myself, I'll be sure to share :).

elucash:
JFYI released in 2.3.4

Stephan202:
:+1: Upgraded and it works. Thanks!

PR #459, opened 2016-10-08
Provide a better experience to Eclipse users

Stephan202:
This is a non-functional change that should make it more attractive for users of Eclipse to contribute to this project. As the `pom.xml` comment explains, this change makes explicit that this project heavily relies on two additional build source paths:

* `${project.build.directory}/generated-sources/annotations`
* `${project.build.directory}/generated-test-sources/test-annotations`

By explicitly listing these using the `build-helper-maven-plugin` they'll automatically be picked up by modern versions of Eclipse, so that the user doesn't have to spend several minutes manually adding these directories to the build path. Tested with the latest version of Eclipse and M2Eclipse (4.6.1 and 1.70, respectively).

NB: Perhaps this trick should also be documented on https://immutables.github.io/?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,6 +151,40 @@
           </compilerArguments>
         </configuration>
       </plugin>
+      <plugin>
+        <!-- This plugin isn't strictly necessary; without it the generated
+        sources are compiled just fine. It's here to make importing the project
+        into Eclipse using M2Eclipse a much smoother process. Thanks to this
+        configuration users won't have to manually add each generated source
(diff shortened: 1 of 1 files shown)


elucash:
I'm Eclipse user myself, but somehow used to manually add directories, or configure Eclipse annotaion processing with snapshot processor or just work with half broken build. Thank you @Stephan202 guess this would be useful for many people, at least for me ) !

Stephan202:
You're welcome :)

Issue #333, opened 2016-05-20
Deep analysis of immutable types should convert immutables-generated contents of guava immutable collections


marquiswang:
After #314, when ```@Value.Style(deepImmutablesDetection = true)``` is enabled , the constructor of @Value.Immutable implementation classes calls ```ImmutableFoo.copyOf()``` on fields that have @Value.Modifiable versions, but the elements of ```ImmutableList``` or other immutable collection fields do not get the same treatment.

Would it be possible, if we had a

```java
@Value.Style(deepImmutablesDetection = true)
@Value.Immutable
@Value.Modifiable
public interface Foo {
    List<Bar> getBars();

    // ...
}


@Value.Immutable
@Value.Modifiable
public interface Bar {
    // ...
}
```

For the ```ImmutableFoo.Builder#addBars``` and ```ImmutableFoo.Builder#addAllBars``` to call ```ImmutableBar.copyOf``` on the arguments as well?

marquiswang:
Same issue with ```ModifiableFoo#from``` it should convert the contents of Immutable collection fields to Modifiable versions.

elucash:
This is known limitation (as in javadoc `This have no effect on collection/container attributes to not interfere with invariant generic types.`). But we can remove the limitation, at least partially, we cannot have a covariant return `ImmutableList<ImmutableValue>` for `List<Value>` but we still can ensure that elements will be converted at runtime. The same for Modifiables.

elucash:
Released in 2.3.7
It's hard for me to tell if it's enough here, I'll close this and let #462 track any remaining functionality
Thank you!

PR #463, opened 2016-10-13
Call copyOf on add methods if deepImmutablesDetection is enabled

marquiswang:
Fix for #333.

In situations like this:

```
    @Value.Style(deepImmutablesDetection = true)
    @Value.Immutable
    @Value.Modifiable
    public interface Foo {
        List<Bar> getBars();

        // ...
    }

    @Value.Immutable
    @Value.Modifiable
    public interface Bar {
        // ...
    }
```
When the following methods are called:
 `ImmutableFoo.Builder#bars(Iterable<Bar>)`,
 `ImmutableFoo.Builder#addBar(Bar)`,
 `ImmutableFoo.Builder#addBar(Bar ...)`,
 `ImmutableFoo.Builder#addAllBar(Iterable<Bar>)`,
 `ImmutableFoo.Builder#from(Foo)`

then the Bar parameters will be converted with ImmutableBar.copyOf
before being added to the list. This is more consistent with
non-collection @Value fields when deepImmutablesDetection is on, and
prevents cases where you have an ImmutableFoo with ModifiableBars
inside.

### DIFF
--- a/value-fixture/src/org/immutables/fixture/deep/Canvas.java
+++ b/value-fixture/src/org/immutables/fixture/deep/Canvas.java
@@ -21,6 +21,7 @@ import org.immutables.value.Value;
 @Value.Style(deepImmutablesDetection = true, depluralize = true)
 public interface Canvas {
 
+  @Value.Modifiable
   @Value.Immutable
   public interface Color {
     @Value.Parameter
@@ -33,6 +34,7 @@ public interface Canvas {
     double blue();
   }
 
+  @Value.Modifiable
   @Value.Immutable
   public interface Line {
     List<Point> points();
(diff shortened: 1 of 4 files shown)


marquiswang:
I've got some issues I wasn't super sure about with this commit. The changes in Immutables.generator are a little bit weird looking and lead to redundant parentheses in the generated code in some cases.

I added `hasAttributeValue` to `ValueAttribute`, which seemed like a good proxy for whether `deepImmutablesDetection` is turned on, but I'm not sure if there is something better, or if the method should be named differently.

elucash:
Thank you @marquiswang for the PR, good job indeed!
Let me review this once again later today, maybe some I'll figure out additional details.

Issue #464, opened 2016-10-14
On the topic of JSON serialization roundtrips


Stephan202:
I'd like to start the discussion with a test class; tried to keep it concise:

```java
package com.example;

import static org.testng.Assert.assertEquals;

import com.fasterxml.jackson.annotation.JsonAutoDetect;
import com.fasterxml.jackson.annotation.JsonProperty;
import com.fasterxml.jackson.annotation.PropertyAccessor;
import com.fasterxml.jackson.databind.ObjectMapper;
import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
import com.fasterxml.jackson.databind.annotation.JsonSerialize;
import java.io.IOException;
import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;
import org.immutables.value.Value;
import org.testng.annotations.Test;

@Test
public final class ImmutablesTest {
    @JsonSerialize
    @JsonDeserialize
    @Value.Immutable
    public abstract static class Dummy {
        @Value.Parameter
        public abstract boolean isSomeProperty();
    }

    @JsonSerialize
    @JsonDeserialize
    @Value.Immutable
    public abstract static class CustomDummy {
        @Value.Parameter
        @JsonProperty("custom_name")
        public abstract boolean isSomeProperty();
    }

    @JsonSerialize
    @JsonDeserialize
    @Retention(RetentionPolicy.CLASS)
    @Target({ElementType.PACKAGE, ElementType.TYPE})
    public @interface Model {}

    @Model
    @Value.Immutable
    public abstract static class DummyWithMetaAnnotation {
        @Value.Parameter
        public abstract boolean isSomeProperty();
    }

    @Model
    @Value.Immutable
    public abstract static class CustomDummyWithMetaAnnotation {
        @Value.Parameter
        @JsonProperty("custom_name")
        public abstract boolean isSomeProperty();
    }

    public void testDummyWithDefaultObjectMapper() throws IOException {
        verifyRoundTrip(getMapper(false), ImmutableDummy.of(true));
    }

    public void testDummyWithCustomObjectMapper() throws IOException {
        verifyRoundTrip(getMapper(true), ImmutableDummy.of(true));
    }

    public void testCustomDummyWithDefaultObjectMapper() throws IOException {
        verifyRoundTrip(getMapper(false), ImmutableCustomDummy.of(true));
    }

[...]

PR #466, opened 2016-10-17
When generating `@JsonProperty`s, mark fields `@JsonIgnore`d

Stephan202:
This is a draft PR following our discussion in #464. Open questions:
- In `Immutables.generator` there's a call to the `declareFields` template in `Renderers.generator`. For some reason I couldn't reference `a.containingType.generateJacksonProperties` for the attributes passed in there. (It says `org.immutables.value.processor.encode.Type.generateJacksonProperties` is unresolvable, which is true.) Do you know how to get that to work (and whether it's important)?
- Did I miss any (other) fields?


elucash:
@Stephan202 thank you for the PR! I'll take a look at `declareFields` in `Renderers.generator`. I cannot immediately notice any missing places, but they may be discovered later.

Stephan202:
Thanks! If I find any, will let you know ;)

Issue #471, opened 2016-10-19
deepImmutablesDetection: 1. broken with optional (bug), 2. no interop with non @Immutable immutables (improvement), 3. documentation missing


vorburger:
Based on a more general version of what I was going on about in https://groups.google.com/forum/#!topic/immutables/TXkXlXEcn4g, how about generating _alternative setters with automated inlined versions for any types having constructor/s?_  So given:

```
@Immutable
interface class Something {
    Thing thing();
}
```

IFF the `Thing` class (or if its an interface or abstract class then its its Immutable\* implementation gen. by Immutables) has a constructor _(or an Immutables `of()` like static factory method; hell even if it had several of either of 'em...)_ then `ImmutableSomething`, in addition to the existing standard `Builder thing(Thing thing)` setter would have a _(several)_ `Builder thing(<Thing cstr args>)` with an implementation body which does `new Thing(<Thing cstr args from setter args>)` (or `Thing.of(..)` - you know what I mean).

See https://groups.google.com/forum/#!topic/immutables/TXkXlXEcn4g for more background re. the idea.

That would be so cool. I think.

elucash:
Hi @vorburger Sorry I didn't have time to make a proper answer on a mailing list. But I just want to let you know that this/similar feature is already implemented: use `deepImmutablesDetection=true` style attribute and your other objects has to have constructors generated, so the builder will have shortcut init methods.

https://github.com/immutables/immutables/blob/master/value-fixture/src/org/immutables/fixture/deep/Canvas.java

```java
@Value.Style(deepImmutablesDetection = true, depluralize = true)
public interface Canvas {

  @Value.Modifiable
  @Value.Immutable
  public interface Color {
    @Value.Parameter
    double red();

    @Value.Parameter
    double green();

    @Value.Parameter
    double blue();
  }

  @Value.Modifiable
  @Value.Immutable
  public interface Line {
    List<Point> points();

    Color color();
  }

  @Value.Modifiable
  @Value.Immutable
  public interface Point {
    @Value.Parameter
    int x();

    @Value.Parameter
    int y();
  }

  default void use() {
    ImmutableLine line = ImmutableLine.builder()
        .color(0.9, 0.7, 0.4)
        .addPoint(1, 2)
        .addPoint(2, 3)
        .addPoint(4, 5)
        .build();

    ImmutableColor color = line.color();

    ImmutableLine.builder()
        .addAllPoints(line.points())
        .color(color)
        .build();
  }
}
```

[...]

PR #472, opened 2016-10-19
Fix minor spelling issues in deepImmutablesDetection() JavaDoc

vorburger:
@elucash it occured to me that instead of smart ass-ing _("sic!")_ in https://github.com/immutables/immutables/issues/471 re. deepImmutablesDetection's JavaDoc I could just make a minor contribution to fix it up.. ;-)


elucash:
thank you!

PR #473, opened 2016-10-19
Call Modifiable factory on elements in add methods if deepImmutablesDetection is enabled

marquiswang:
Like #333 but for Modifiables instead of Immutables...

In situations like this:

    @Value.Style(deepImmutablesDetection = true)
    @Value.Immutable
    @Value.Modifiable
    public interface Foo {
        List<Bar> getBars();

        // ...
    }

    @Value.Immutable
    @Value.Modifiable
    public interface Bar {
        // ...
    }

When the following methods are called:
 ModifiableFoo#setBars(Iterable<Bar>),
 ModifiableFoo#addBar(Bar),
 ModifiableFoo#addBar(Bar ...),
 ModifiableFoo#addAllBar(Iterable<Bar>)
 ModifiableFoo#from(Foo)

then the Bar parameters will be converted with

ModifiablePoint.create().from()

before being added to the list if they are not already ModifiablePoints.
This is more consistent with non-collection @Value fields when
deepImmutablesDetection is on, and prevents cases where you have an
ModifiableFoo with ImmutableBars inside.

### DIFF
--- a/value-fixture/test/org/immutables/fixture/deep/DeepImmutablesDetectionTest.java
+++ b/value-fixture/test/org/immutables/fixture/deep/DeepImmutablesDetectionTest.java
@@ -19,6 +19,7 @@ import com.google.common.collect.ImmutableList;
 import org.immutables.fixture.deep.Canvas.Line;
 import org.junit.Test;
 
+import java.util.ArrayList;
 import java.util.Collections;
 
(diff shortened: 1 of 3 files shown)


elucash:
Reviewed, makes sense, Thank you for PR!

Issue #492, opened 2016-11-20
Warnings in hashCode generated code with using ErrorProne tool


idyatlov:
When Immutables used and compiled with [ErrorProne](https://github.com/google/error-prone), using [Gradle ErrorProne plugin](https://github.com/tbroyer/gradle-errorprone-plugin), there is a warning [BoxedPrimitiveConstructor](http://errorprone.info/bugpattern/BoxedPrimitiveConstructor) when using double values. In it said that we should use Double.valueOf() insyead of new Double() because of allocations.

elucash:
Thank you for the PR! This will be released in the next version

elucash:
released in 2.4.0, thank you!

PR #493, opened 2016-11-21
use Double.valueOf() instead of new Double()

idyatlov:
Fixed Issue #492


elucash:
Thank you for the PR!

Issue #496, opened 2016-11-25
Possible to generate an implementation that allows nulls?


renedewaele:
I often run into the following issue. I want to define interfaces without `@Nullable` or optionals present for the end state of an object to make sure no required attribute is null. However, while building up this object (over the course of hours or sometimes days) I don't mind that some of its fields are null. I.e. I only want the not-null check to happen when the object is completely built.

I don't want to use a Builder for this because the Builder object is not meant to be queried or persisted and is mutable so inherently unsafe to use in events etc.

Basically, if specified by annotation it would be neat if the framework would generate an implementation of my interface that allows nulls and can be converted to the Immutable version (with null-checks) on demand. Note that the generated class may contain object attributes that also need to be converted to their Immutable version.

What do you think?

renedewaele:
Another approach would be to add a style that doesn't force null-checks while the object is created but waits until a method (e.g. `isValid()`) is invoked, ideally also checking all attributes that have this style enabled.

elucash:
@renedewaele Have you tried to use `@Value.Modifiable` for this? I somehow reluctant to gradually relax consistency and other guarantees in order to turn immutable value object into mutable, inconsistent, "anything can be null" kind of beans.
But modifiable companion objects are a little bit different, they are a sort of the builder or buffer (Like `ArrayList` for `ImmutableList`). Can have some attributes not initialized, but still non-null in its definition, then you can `toImmutable` and everything will be checked as one would expect.

garakelian:
I agree completely with @renedewaele.  This is an essential feature for us, and I've just forked the repository to see if I can figure out how to disable this altogether (while maintaining immutability). Let me give you a use case, and try to convince you...

For our REST API, we use beans that look like this one for incoming requests:

`@ApiModel
	@Value.Immutable
	@JsonSerialize(as = ImmutableChangePasswordRequest.class)
	@JsonDeserialize(builder = ImmutableChangePasswordRequest.Builder.class)
	public static interface ChangePasswordRequest extends RestRequest {
		@NotNull
		@Size(min = 1, max = ValidationConstants.MAX_PASSWORD_LENGTH)
		@ApiModelProperty(value = "Old password", required = true, example = "0ldPassw0rd!", position = 100)
		@Nullable
		public String getOldPassword();

		@NotNull
		@Size(min = 1, max = ValidationConstants.MAX_PASSWORD_LENGTH)
[...]

Issue #26, opened 2014-09-07
validation: support for javax.validation API


toellrich:
I came across this awesome library when you posted that message on the Guava mailing list the other day. Too bad they haven't answered you yet. 

I would like to start using it for generating objects that are exchanged via JAX-RS in combination with the bean validation API for which Jersey has build-in support. Unfortunately that requires me to follow the JavaBean convention of prefixing the method names with 'get'/'is':

http://docs.jboss.org/hibernate/validator/5.1/reference/en-US/html_single/#d0e349

Would you consider prefixing the generated builder method names with 'set' if the method names start with 'get'/'is'? Setters with getter method names in a builder, that's just wrong.

I understand where you're coming from in eschewing this convention in your documentation (after all if there are no setters why should there by getters), but maybe in this case an exception could be made?

Anyway, thanks for sharing this great library.

DavidLeoni:
Hi,

regarding validation api support, I noticed that when `@Valid` annotation is added to a getter, the example below throws on exception upon validation because `@Valid` tag is copied to generated Immutable class getter and this upsets standard hibernate implementation (tried hibernate-validator 5.1.3.Final and Immutables 2.0.16):

```
class Person {
    
}

@Value.Immutable
abstract class Car {
    @Valid
    public abstract Person getDriver();
}

public class TestGen {

    @Test
    public void testValidate() {        
        ValidatorFactory factory = Validation.buildDefaultValidatorFactory();
        Validator validator = factory.getValidator();
         Set<ConstraintViolation<ImmutableCar>> constraintViolations
                = validator.validate(ImmutableCar.builder().driver(new Person()).build());
    }

}

```

This is the thrown exception:

```
javax.validation.ConstraintDeclarationException: HV000131: A method return value must not be marked for cascaded validation more than once in a class hierarchy, but the following two methods are marked as such: public abstract it.davidleoni.experiments.immutables.test.Person it.davidleoni.experiments.immutables.test.Car.getDriver(), public it.davidleoni.experiments.immutables.test.Person it.davidleoni.experiments.immutables.test.ImmutableCar.getDriver(). 
```

An ad-hoc solution might be not to copy `@Valid` annotation, or a generic solution might introduce a system to prevent copying of certain user-specified annotations.

augustotravillio:
Hi David
There's more generic solution for this, there's `Style.passAnnotations`, but it had to be fixed/expanded to work on attribute level annotations also (see #139).
[...]

PR #513, opened 2016-12-27
Add style which allows developer to configure how validation is performed, including new option to disable null checks, and another option to use Java Validation API

garakelian:
@elucash, this is a very rough, untested, first cut at disabling required fields, per our discussion in #496. Just wanted you to take a look at it, and let me know if I'm on the right path.

The intent of this PR is disable checks for "required" attributes, such that:
1. You don't have to add "@Nullable" to every field
2. You don't have to provide default values for primitives if you are content with 0 and false.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -1766,8 +1766,24 @@ checkNotIsSet([v.names.isSet](), "[v.names.raw]");
     checkRequiredAttributes();
     [/if]
   [/if]
+[if type.useJavaValidationApi]
+    [returnType] bean = _[type.names.build]();
+    javax.validation.Validator validator = javax.validation.Validation.buildDefaultValidatorFactory().getValidator();
+    java.util.Set<javax.validation.ConstraintViolation<[returnType]>> constraintViolations = validator.validate(bean);
+    if (!constraintViolations.isEmpty()) {
+      throw new javax.validation.ConstraintViolationException(constraintViolations);
+    }    
+    return bean;
+[else]  
+    [builderReturnValue type]
+[/if]
+  }
+[if type.useJavaValidationApi]
+  private [returnType] _[type.names.build]() {
     [builderReturnValue type]
   }
+[/if]
+  
   [if type.isGenerateBuildOrThrow]
   [if classpath.isJava8 or (not type.generateJdkOnly)]
 
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -705,7 +705,17 @@ public [type.typeAbstract.relative] [v.names.with]([v.atNullability][v.type] val
 [/if]
 [if type.useValidation]
 
+[if type.useJavaValidationApi]
+  private static final javax.validation.Validator [disambiguateField type 'validator'] = javax.validation.Validation.buildDefaultValidatorFactory().getValidator();
+[/if]
+
   private static [type.generics.spaceAfter][type.typeImmutable.relative] validate([type.typeImmutable.relative] instance) {
+[if type.useJavaValidationApi]
+    java.util.Set<javax.validation.ConstraintViolation<[type.typeImmutable.relative]>> constraintViolations = [disambiguateField type 'validator'].validate(instance);
+    if (!constraintViolations.isEmpty()) {
+      throw new javax.validation.ConstraintViolationException(constraintViolations);
+    }
+[/if]
 [for vm in type.validationMethods]
   [if vm.normalize]
     instance = ([type.typeImmutable.relative]) instance.[vm.name]();
@@ -1020,9 +1030,6 @@ public interface BuildFinal[type.generics] {
 [if type.constitution.outsideBuilder]
   [rr.staticFields setters]
   [rr.staticMethods setters]
-[/if]
-[if type.useJavaValidationApi]
-  private static final javax.validation.Validator [disambiguateField type 'validator'] = javax.validation.Validation.buildDefaultValidatorFactory().getValidator();
 [/if]
   [for m in mandatories, BitPosition pos = positions m]
   private static final long INIT_BIT_[toConstant m.name] = [literal.hex pos.mask];
@@ -1769,23 +1776,8 @@ checkNotIsSet([v.names.isSet](), "[v.names.raw]");
     checkRequiredAttributes();
     [/if]
   [/if]
-[if type.useJavaValidationApi]
-    [returnType] bean = [type.names.build]Unvalidated();
-    java.util.Set<javax.validation.ConstraintViolation<[returnType]>> constraintViolations = [disambiguateField type 'validator'].validate(bean);
-    if (!constraintViolations.isEmpty()) {
-      throw new javax.validation.ConstraintViolationException(constraintViolations);
-    }    
-    return bean;
-[else]  
-    [builderReturnValue type]
-[/if]
-  }
-[if type.useJavaValidationApi]
-
-  private [returnType] [type.names.build]Unvalidated() {
     [builderReturnValue type]
   }
-[/if]
   [if type.isGenerateBuildOrThrow]
   [if classpath.isJava8 or (not type.generateJdkOnly)]
 
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -1021,6 +1021,9 @@ public interface BuildFinal[type.generics] {
   [rr.staticFields setters]
   [rr.staticMethods setters]
 [/if]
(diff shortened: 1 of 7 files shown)


garakelian:
@elucash, I tested this locally, and it works well.  When this style is enabled, ValueAttribute.isMandatory() always returns false, which results in the initBits stuff never getting generated.

garakelian:
@elucash - So, I updated this PR to add support for Java Validation API. There are now 3 options for validation: (1) turn off, e.g. the original intent of this PR (2) simple, which is what the project currently supports, and (3) use Java Validation API to verify bean, replacing current approach.

elucash:
@garakelian I very like your work, and how you've managed to get thru the stuff, the use of `isMandatory` perfectly matches the intent!

I do think that dedicated Validation API support is, probably, a bit redundant, you probably know about BYOM approach I outlined in #26. There's a problem with validator management: validator factory/validator creation is potentially CPU and memory expensive. When using BYOM, you can decide if you want a single static validator or a passed in request-/app-scoped instance. Putting aside performance concerns, I think it could be nice to have easy, switch-on Validation API support. But it seems wasteful to have it only for the experimentation reasons, switching to BYOM-like approach later.
My thumbs up for the initial disable-mandatory-check flag approach.
Validator instance handling is currently a stumbling block for me, convenience-oriented experimentation-only functionality is something we would like to avoid. If we can solve the validator instance handling somehow, there remaining suggestion would be to not use validation on a builder path, but rather use precondition checking phase (where we invoke checker methods, interning and canonicalization hooks), the reason is that validation should work consistently across all construction paths, not only via builder.

Let's keep in mind that we're also want to avoid introducing mandatory runtime dependencies, at least in the default module. What would I propose? We can create separate API module like `org.immutables:validation` and place mix-in there. The mix-in would have a single static validator created for straightforwardness. We can introduce a validation annotation in that module to auto-implement the mix-in as the alternative to inheriting it. Actual supporting annotation processing code would still be inside `value-processor` module, but we will have cleaner/optional runtime API which itself will have a dependency on `javax.validation` API

garakelian:
@elucash Thanks for the feedback. I honestly had not read #26 before. After my first implementation, using `isMandatory`, I began to think about it some more and in a moment of inspiration, just made the change. I like that your framework's default behavior is to try to ensure a bean is properly constructed, and upon reflection, I thought I should do more than simply disable this feature. Thus, I added support Validation API.

With regards to the idea of a mix-in, let me say, I completely agree that we don't want to introduce any runtime dependencies, which is why I'm hesitant to agree to that approach. One of the coolest features of code generation within a Java annotation processor is that it has access to the classpath of the enclosing code!  When I first used Immutables, I was very impressed to see that you checked if Guava was on the classpath, and you modified the code generation appropriately. I'd like to do the same thing here if possible.

However, I understand your concern about initializing the ValidatorFactory multiple times. I'm less concerned however, with supporting generic implementations of "validation", because it's a complex issue, and because the Hibernate Validator is the reference implementation of the JSR. Spring uses it, issue #26 references it, etc. It's like Guava - widely used and well tested. If we really wanted to that later on, the BuilderValidationMethod is a enum, and we could add another option there.

So, how can we solve the multiple initializations? A simple approach would be to use a static field to hold a reference to the Validator. From the Hibernate Validator documentation (http://docs.jboss.org/hibernate/validator-backup/4.0.1/reference/en-US/html_single/): "A Validator instance is thread-safe and may be reused multiple times". This would be a trivial change.

Another approach, also not too difficult, would be to add a Style attribute to allow the developer to specify a class instance that would contain a static factory method, something like getValidatorFactory() that returns javax.validation.ValidatorFactory.

elucash:
The usage of a static singleton is ok in general. The caveat is that there will be validator per each value class. This is a drastic improvement from having validators created over and over again for each value object instantiation but still can be suboptimal. This should be acceptable, though. (with a separate mix-in using static field, all inheriting objects are using that one instance)

The other thing which worries me is that validation should be executed thru all construction paths. You really should try to move the validation logic to the "centralized" place, where each object will go thru validation regardless if it was constructed using constructor or intented. Here's the points in the code involved:

* `ValueType.isUseValidation()` should return true early if jsr303 should be used.
*  There's validation sequence. Look for the following fragment:
```
[if type.useValidation]

  private static [type.generics.spaceAfter][type.typeImmutable.relative] validate([type.typeImmutable.relative] instance) {
[for vm in type.validationMethods]
  [if vm.normalize]
    instance = ([type.typeImmutable.relative]) instance.[vm.name]();
  [else]
    instance.[vm.name]();
  [/if]
[/for]
[if type.useSingletonOnly]
    return INSTANCE != null ? INSTANCE : instance;[-- may be null during initialization --]
[else if type.generateOrdinalValue]
    return ([type.typeImmutable.relative]) instance.domain.internOrdinal(new InternProxy(instance));
[else if type.useInterned]
  [if type.generateJdkOnly]
    [atNullable][type.typeImmutable.relative] interned = INTERNER.putIfAbsent(new InternProxy(instance), instance);
    return interned != null ? interned : instance;
  [else]
    return INTERNER.intern(new InternProxy(instance)).instance;
  [/if]
[else if type.useSingleton]
    return INSTANCE != null && INSTANCE.equalTo(instance) ? INSTANCE : instance;[-- may be null during initialization --]
[else]
    return instance;
[/if]
  }
[/if]
```
* The jsr303 fragment should be inserted into the very beginning of the validation routine, i.e. before `[for vm in type.validationMethods]`

Please, do not consider this nitpicking, I just want to align this functionality with existing features and overall design. I do really appreciate your efforts to make the library better! Thank you!

garakelian:
@elucash Thanks again for the feedback. Good comments. I definitely didn't intend to leave out any code paths that could escape validation. I'll look for the validation sequence you suggested in the morning and update the PR. Happy to help here, thanks for your excellent library!

Stephan202:
The changes now under discussion seem to assume that the default `Validator` (as acquired through `Validation.buildDefaultValidatorFactory().getValidator()`) will suffice. It's easy to envision a new feature request in the not-so-distant future asking for a custom `Validator` to be used. For example, a user may want to create a `Validator` instance in an environment in which no EL implementation is present (like in [HV-1131](https://hibernate.atlassian.net/browse/HV-1131)).

So it seems to me now is the time to figure out how to make this customizable. Since I also think statics should be avoided: how about some kind of "Immutables factory"? Sketch:

```java
MyDtoFactory factory = MyDtoFactory.create(someValidator);
SomeDtoType dto = factory.someDtoTypeBuilder()......build();
```

The idea here would be that:
1. `MyDtoFactory` would be generated. Its name is customizable, and it allows the construction of all immutables within the same "group", as configured using `@Style`.
2. `MyDtoFactory` would have one or more methods for each immutable type it knows about, corresponding to the creator and builder methods currently defined on the immutable types themselves.
3. `MyDtoFactory` is parametric on a `Validator`, allowing customization thereof. As a big extra plus, it obviates the need for a static `Validator` instance per immutable type; now there's just one instance for all types.

(Perhaps my proposal in point 1 is not possible due to the way annotation processing works, esp. in the context of incremental compilation... but I hope not. In that case there'd have to be one factory per immutable type, which is less nice.)

WDYT?

garakelian:
@Stephan202 I agree that we'll eventually need a way to create a custom validator, but I don't think the best way to do that is using a singleton factory as you describe. First, it's tedious, as most projects will likely have lots of generated classes, and now you've got to create singletons for all of them, simply so you can pass the Validator factory to it.  Second, what about Jackson deserialization? Although Jackson supports deserialization through a builder, does it support a singleton builder?

I personally see this feature working the other way around. I think the builder should reach out (using a Style attribute described above), to get a static ValidatorFactory, which would enable a developer to cache/customize the creation anyway they see fit.

garakelian:
@Stephan202 and @elucash - After some research, I don't think we need to do anything regarding the installation of custom JSR 303 validators. This is part of JSR303 itself.

This is code used to initialize the validator:
```
  private static final Validator validator = Validation.buildDefaultValidatorFactory().getValidator();
```

If you look at the code for Validation.buildDefaultValidatorFactory, you can see that it uses standard Java service discovery mechanisms to [find the implementation](http://grepcode.com/file/repo1.maven.org/maven2/javax.validation/validation-api/1.1.0.Final/javax/validation/ValidationProviderResolver.java#ValidationProviderResolver.getValidationProviders%28%29).

elucash:
@Stephan202 as I'm concerned about custom validator creation or propagation, it should be handled in DYI manner in a way described in #26 using custom mix-in. The functionality here is similar to strong interning we provide: default or nothing. I'm against introducing fine-grained configuration to something that can be easily built on top. The only exception is simple out of the box experience for some standard (de-facto or de-jsr-ure). I'm planning to highlight how jsr303 can be integrated/configured in the javadoc for the new switch and the guide.

elucash:
@garakelian merged, thank you! Let me muse a little on the annotation API itself before releasing. I have a couple of thought to confirm with you. I'll return to this in two (hope not more that this). Thank you!

Issue #516, opened 2017-01-03
Extend `@NaturalOrder` support to other collection types


Stephan202:
The `@NaturalOrder` annotation currently applies to `SortedSet`s, `SortedMap`s, and (a number of?) their subtypes. If the annotation is applied to a `List`, say, then it has no effect.

I'd like to propose that support is also added for other types, in cases where this makes sense. Examples:
* `List`.
* `Map` (through use of `LinkedHashMap`).
* Guava's `Immutable{Set,List}Multimap` (the builders accept comparators).

I imagine this may not be trivial to add. Before I attempt to dive in: do you agree with this proposal in principle?

elucash:
@Stephan202 Thank you for the proposal. I don't think this functionality is necessary, mostly because it makes things needlessly complicated from the standpoint of semantics. If you declare list - it imposes insertion ordering. When you have Map, well, a map does not guarantee any ordering, but an immutable implementation will also follow insertion ordering (unmodifiable `LinkedHashMap` or `ImmutableMap`). Use `Sorted/NavigableSet/Map` to keep the content auto-sorted, because of the nature of the sorted collections, not because there's an annotation. `@NaturalOrder` and `@ReverseOrder` are simply to make it explicit what comparator will be used. If you need a custom comparator, you're also "out of luck": just create `SortedSet` fully outside the builder.
I know there's always a tension to sweep as much as possible under the rug of the code generation and have no boilerplate with only handful of fancy annotations on top (sorry, I'm guilty in this too). But here is exactly the case where explicit code would work best. You either create and sort list before setting onto builder, or you can use sorted views using `@Value.Derived` attribute. There's a recent post on StackOverflow with the similar example: http://stackoverflow.com/questions/41376633/how-would-it-be-advised-to-get-immutables-to-build-this-object-with-this-sort/41378182#41378182

elucash:
Although it would be really great if someone helped to get `SortedMultiset` and `SortedMultimap` supported via  `@NaturalOrder` and `@ReverseOrder` in the same manner as `SortedSet` and `SortedMap` are supported. The entry point in the source code would be `AttributeTypeKind` enum, and then some accessor and a number or template changes required for this to work.

Stephan202:
[...]

PR #518, opened 2017-01-04
Support @{Natural,Reverse}Order on SortedMultiset

Stephan202:
From the commit message:
```
While there, also:
- Tweak the error messages in `ValueAttribute#checkOrderAnnotations`. (Note
  that `NavigableSet` extends `SortedSet` and `NavigableMap` extends
  `SortedMap`, so those don't need to be listed explicitly.)
- Fix the (unused) `AttributeTypeKind#isListMultimap`.
```

This PR relates to #516. Note that:
1. I added some tests and also tried the changes with `ImmutableSortedMultiset` in a local project and things seem to work, but I'm not sure I covered all cases.
2. I can trigger the error for when one specifies both `@NaturalOrder` and `@ReverseOrder`, but when I annotate an `ImmutableList` with (just) `@NaturalOrder`, the associated error is not triggered (as I also mentioned in #516). No idea why this is.

### DIFF
--- a/value-fixture/src/org/immutables/fixture/OrderAttributeValue.java
+++ b/value-fixture/src/org/immutables/fixture/OrderAttributeValue.java
@@ -15,6 +15,7 @@
  */
 package org.immutables.fixture;
 
+import com.google.common.collect.SortedMultiset;
 import java.util.NavigableMap;
 import java.util.NavigableSet;
 import java.util.SortedMap;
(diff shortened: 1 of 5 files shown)


elucash:
You've followed all the patterns and anti-patterns we have in the codebase, so this is very organic addition! Brilliant work! Thank you @Stephan202

Stephan202:
Thanks! :bowing_man:

Issue #525, opened 2017-01-15
@NaturalOrder on a collection of immutable generated types


Stephan202:
It seems I hit a limitation of the sanity check performed for the `@NaturalOrder` and `@ReverseOrder` annotations. Consider the following class:

```java
@Value.Immutable
@Value.Style(typeAbstract = "Abstract*", typeImmutable = "*")
public abstract class AbstractSortedMultisetWrapper {
    @Value.NaturalOrder // Works.
    public abstract ImmutableSortedMultiset<AbstractElem> getAbstractElems();

    @Value.NaturalOrder // Causes error.
    public abstract ImmutableSortedMultiset<Elem> getElems();

    // The problem is/may also be exhibited if the element type is declared
    // in a file of its own.
    @Value.Immutable
    public static abstract class AbstractElem implements Comparable<AbstractElem> {
        public abstract int getValue();

        @Override
        public final int compareTo(AbstractElem other) {
            return getValue() - other.getValue();
        }
    }
}
```

This fails to compile, triggering the following error:
```
error: @Value.Natural requires that a (multi)set's elements or a map's key are Comparable
```

Stepping through the code, it seems happens because the `Elem` type isn't known (`TypeKind.ERROR`) at the moment that `ValueAttribute#introspectType()` is invoked. As a result `ValueAttribute#isComparableKey()` yields `false`.

So my question is: is there any way we can make sure the element type is resolved in time, or can we relax the check such that it doesn't complain about unresolvable types?

elucash:
probably, the check can be disabled if a type is unresolved, there is no way to analyze not-yet-generated types

Stephan202:
Thanks for the super-fast response! I had already started an investigation along those lines; the result is in PR #526.

Stephan202:
I'll close this issue since the PR has been merged. Thanks again :).

elucash:
this was released in 2.4.1. Thank you!

PR #526, opened 2017-01-15
Allow unresolved ordered collection elements

Stephan202:
From the commit message:

```
Relax the check for `@NaturalOrder` and `@ReverseOrder` to allow unresolved
element/key types, as it is not possible to determine with certainty whether
such types are in fact `Comparable`.
```

This is a possible fix for #525. I replaced `#isComparableKey()` with `#isMaybeComparableKey()` because it was used nowhere else. If you prefer I can also retain the original method alongside the new method, or split the logic into a method `#isUnresolvedKey()`. Let me know :)


elucash:
thank you!

Stephan202:
Thank _you_ for merging :)

PR #527, opened 2017-01-15
Assorted fixes

Stephan202:
While reviewing PR #526 once more I noticed that I forgot to rename the test class. So this PR rectifies that and throws in two more tweaks:

1. Rename `SortedMultisetWrapper` -> `SortedCollectionWrapper` to match its content.
2. Deduplicate logic in `ValueAttribute#checkOrderAnnotations`.
3. Fix typo: "a map's key are Comparable" -> "a map's keys are Comparable".


elucash:
LGTM
thanks!

Issue #541, opened 2017-02-02
Modifiable with deepImmutablesDetection in different packages


DennisRippinger:
I discovered that the usage of @Modifiable together with deepImmutablesDetection can lead to compilation problems if the referenced value is in another package. It fails to compile because it generates no import or FQN of such values. I created a [MWE](https://github.com/DennisRippinger/immutables-MWE).

This problem could be located in the [Modifiables.generator](https://github.com/immutables/immutables/blob/master/value-processor/src/org/immutables/value/processor/Modifiables.generator#L720), here typeModifiable.relativeRaw and typeModifiable.relative appear to return just the name, or whatever is needed for the PostprocessingMachine.

elucash:
yep, I think fqn should be used instead of relative. Thank you for reporting this!

elucash:
Thank you for the PR, released in 2.4.3

PR #543, opened 2017-02-03
#541 Modifiable with deepImmutablesDetection in different packages

DennisRippinger:
As mentioned in #541 should the typeModifiable have some kind of FQN value.

This PR adds absolute and absoluteRaw fields to typeModifiable and a simple test to fixture.


elucash:
Thank you! Hope this works (test convinces)

Issue #578, opened 2017-03-20
Modifiable#addAll* misses null check on @Nullable


DennisRippinger:
Assume a modifiable model with `@Nullable List<Object> nullableObjects();` This will create a code similar to:
```
  public ModifiableModel addAllNullableObjects(Iterable<Object> elements) {
    if (nullableObjects == null) {
      nullableObjects = new ArrayList<Object>();
    }
    for (Object element : elements) {
      addNullableObjects(element);
    }
    return this;
  }
```
This will lead to a NPE when the foreach loop accesses a null reference. This is problematic when calling modifiableModel.from(model) which eventually calls addAll*.
Here is a MWE: https://github.com/DennisRippinger/immutables-MWE

elucash:
released in 2.4.6 thank you for PR!

PR #579, opened 2017-03-20
#578 Modifiable#addAll* misses null check on @Nullable

DennisRippinger:
Should fix a NPE when creating a modifiable model from an immutable one if a nullable List is actually null.


elucash:
Thank you for PR!
You know, that was like a bit controversial as null was designed to be ok when `set`ting iterable, not when adding from. But then, it's modifiable and nullable attribute, so let it be the way of your PR.

PR #582, opened 2017-03-21
Support Java 6 runtime when using serial module

trask:
Replacing Java 7 String switch statements (similar to my earlier PR #43).


elucash:
Guess it's ok, structural serialization was always about compatibility/flexibility rather than speed (when I know this also questionable as we've run benchmarks on these switches)
Thank you for PR!

Issue #570, opened 2017-03-08
Numeric comparisons triggering error in errorprone


lehphyro:
The following class generates errorprone errors:
```java
@Value.Immutable
abstract class AbstractTestClass {
    public abstract Optional<BigDecimal> getNumericValue();
}
```
The error message is:
```
error: [NumericEquality] Numeric comparison using reference equality instead of value equality
    if (this.numericValue == newValue) return this;
                               ^
    (see http://errorprone.info/bugpattern/NumericEquality)
  Did you mean 'if (Objects.equals(this.numericValue, newValue)) return this;' or 'if (this.numericValue.equals(newValue)) return this;'?
```

Stephan202:
Since this code is part of a "wither" method, this is not a functional defect, but it'd be nice to avoid the warning regardless. Because [`NumericEquality`](http://errorprone.info/bugpattern/NumericEquality) has severity `ERROR` it cannot simply be suppressed using `-XepDisableWarningsInGeneratedCode`. A workaround is to explicitly downgrade this particular check to `WARN` severity, but that's not a burden that should be placed on all users of both Immutables and Error Prone.

I see two possible solutions:
1. Update the generated code to use `#equals` for subtypes of `java.lang.Number`. This is my preferred solution, but perhaps this is deemed too expensive. (Because in the common case the values are probably not equal.)
2. Add `@SuppressWarnings("NumericEquality")` to "wither" methods accepting an (optional) subtype of `java.lang.Number`.

@elucash, I'm willing to try and open a PR for either approach. Which do you prefer? (Or do you perhaps see a third option?)

Stephan202:
Thinking about this some more, using `#equals` for `java.lang.Number` subtypes seems correct only if those subtypes are immutable. So that would require a whitelisting approach (`BigInteger`, `BigDecimal`, `UnsignedInteger`, `UnsignedLong`...), but not resolve the issue altogether.

So `@SuppressWarnings("NumericEquality")` seems the way to go. I see two potential implementations:
1. Simply add this annotation to the top of the class, where currently by default `@SuppressWarnings({"all"})` is generated. (Note that Error Prone does not accept "all". Maybe it should?)
2. Add custom `@SuppressWarnings` annotations only on the methods that trigger the error.

The first solution is decidedly simpler. WDYT?

elucash:
[...]

PR #583, opened 2017-03-21
Prevent Error-Prone `NumericEquality` warning

Stephan202:
In a1d3518d4851291103fd98a449fbbd8fded90579 triggering `NumericEquality`
warnings is prevented for optional numbers. This change tackles non-optional
`java.lang.Number` instances.

This relates to #570.

### DIFF
--- a/value-fixture/src/org/immutables/fixture/with/ForEquals.java
+++ b/value-fixture/src/org/immutables/fixture/with/ForEquals.java
@@ -1,0 +1,49 @@
+/*
+  Copyright 2017 Immutables Authors and Contributors
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+package org.immutables.fixture.with;
+
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.util.List;
+import java.util.Optional;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public abstract class ForEquals {
+    public abstract int myInt();
+
+    public abstract Optional<Integer> myOptinalInt();
+
+    public abstract List<Integer> myIntList();
+
+    public abstract BigDecimal myBigDecimal();
+
+    public abstract Optional<BigDecimal> myOptinalBigDecimal();
+
+    public abstract List<BigDecimal> myBigDecimalList();
+
+    public abstract RoundingMode myRoundingMode();
+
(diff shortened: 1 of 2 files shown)


Stephan202:
I checked that this change fixes the problem we've observed, but please do verify carefully. Note that this change also syncs the JavaDoc and code-level checks, so that's an extra plus.

elucash:
Thank you for PR!
the change to `hasSimpleScalarElementType` seems mostly equivalent (Albeit it was initially designed for checking containers' element types). Can you please check if having `or v.enumType` along with primitive would be ok wrt error-prone, so we would still `==` enums.
Test value object can check enum attribute and list of enums, to see if we handle it correctly.

Stephan202:
Will do :)

elucash:
Superb! Thank you @Stephan202!

Stephan202:
You're fast haha! I was still writing out some notes:
- I compared the code generated for the `ForEquals` class with what was generated at 657afa84a00d092d9f60ca622cd514b2ea8089c5 (i.e., just before your initial changes). Code-wise only the `BigDecimal` comparisons are changed, as required.
- In terms of Javadoc, there is now one place where it says enums are compared using an "An equals check" rather than a "A shallow reference equality check". I can fix that, but it'd mean that the if/else branches between the code and Javadoc don't line up, while for all intends and purposes the new text isn't wrong.

I do wonder: is there some way to validate/check the generated code, so that the regression I introduced could have been caught at build time?

Stephan202:
Something else I noticed: even before 657afa84a00d092d9f60ca622cd514b2ea8089c5 (and now still; this is not a regression), the following code is generated:
```java
  public final ImmutableForEquals withMyOptinalRoundingMode(RoundingMode value) {
    @Nullable RoundingMode newValue = Preconditions.checkNotNull(value, "myOptinalRoundingMode");
    if (Objects.equal(this.myOptinalRoundingMode, newValue)) return this;
    return new ImmutableForEquals(
        this.myInt,
        this.myOptinalInt,
        this.myIntList,
        this.myBigDecimal,
        this.myOptinalBigDecimal,
        this.myBigDecimalList,
        this.myRoundingMode,
        newValue,
        this.myRoundingModeList,
        this.myObject,
        this.myOptinalObject,
        this.myObjectList);
  }
```

That `Objects.equal` could be replaced with an `==`. Maybe for another PR :). (But if you immediately see how to fix it: go ahead.)

elucash:
> I do wonder: is there some way to validate/check the generated code, so that the regression I introduced could have been caught at build time?

I always wanted to avoid testing literal code output as it overspecifies code, making test very fragile for any changes including cosmetic ones. Usually, if you generate some code, the assertion could be the successful compilation, where otherwise it would not compile (sample code references generated elements and expect) or runtime test can prove that code is properly generated. It is more complicated with warnings, as actually stuff like google's compile-testing can help with that. But I've yet to actually use it, so now I inspect warnings as I run builds.'

> That Objects.equal could be replaced with an ==. Maybe for another PR :). (But if you immediately see how to fix it: go ahead.)

It can be different branch of `with` method generation. Will see

Stephan202:
Thanks for the explanation :)

I think I found the place where to change the code, and I also noticed I made a typo ("Optinal"). WIll try to open an extra PR shortly.

Stephan202:
For those reading along at home: the new PR is #584.

PR #584, opened 2017-03-21
Followup of #583

Stephan202:
Summary of changes:
- Fix typo in `org.immutables.fixture.with.ForEquals`.
- Compare enums by reference.

(Looking again at the code derived from `ForEquals`, only the equality check mentioned in https://github.com/immutables/immutables/pull/583#issuecomment-287989754 is changed, as desired.)


elucash:
Hope we're done for now, thank you for the efforts!

PR #601, opened 2017-04-19
Allow Metainf.Service annotation on static nested classes

auke-:


elucash:
Thank you for the improvement! The PR have tests failing will see what's happening

elucash:
put the tests to other (value-fixture) module to make sure tests are compiled against current version, so the build is fixed for now. Well, there are some troubles with self-bootstrapping )

auke-:
Thanks! The test will succeed when a new release is made (and the released version contains the change)

Issue #606, opened 2017-04-29
GsonMessageBodyProvider fails to serialize Boolean


dgrant:
I have a map like this:

```java
@Gson.TypeAdapters
@Value.Immutable
public interface MapTest {
  Map<String, Object> mapObject();
}
```

And if I do
```java
    return ImmutableMapTest.builder()
            .putMapObject("object", true)
            .build();
```

it fails serialize the boolean value. I get an exception in JsonGeneratorWriter.

I will submit a PR with the test case I created that causes the bug.

elucash:
Thank you for raising this! This is a hard topic because it's not obvious for me what is the expected behavior. After checking the generated code everything "looks good". As the declared value type is the `Object` neither Immutables, nor Gson have any idea how to serialize/deserialize the `Object`. In theory, it should be possible to register special type adapter with Gson for `Object` class and have there some runtime type/JSON type analysis to handle types one can think could be handled automatically, i.e. auto-convert JSON structures to `List<Object>, Map<String,Object>, String, Boolean, Number, nulls` and back. Currently,  I lean on thinking that it is out of scope for Immutables to provide such generic handlers. Just to mention, Gson have so called "tree API" comprised of `JsonElement` and its subclasses, while it's not the same, but Gson provides type adapter for this automatically, so declaring `Map<String, JsonElement>` you can get any JSON element/object/array out of value.

dgrant:
Updated from one of our engineers:

Setting this flag to true seems to mess with the serialization into JSON.

With this flag, Boolean objects (in our case, the boolean primitives are being put into a Map (and converted into Boolean objects) are invalid values, and the JSON string produced just ends
ex:
``` 
{
    "someReallyCoolObject" : {
        "thisDoubleIsOnFire" : 1.0,
        "aBooleanWithSwag"
```  
when it should be:
``` 
{
    "someReallyCoolObject" : {
        "thisDoubleIsOnFire" : 1.0,
        "aBooleanWithSwag" : true,
        "thisBooleanIsLit" : false
    }
}
``` 

For some strange reason, disallowing Jackson solves this issue.

elucash:
I agree that it's not the only issue, have to run and see what actually happens

augustotravillio:
[...]

PR #607, opened 2017-04-29
GsonMessageBodyProvider unable to serialize Boolean

dgrant:
Test reproduces the bug #606


dgrant:
FYI: the test is supposed to fail, the test exposes a bug. I've tried investigating it but haven't found the source of the problem yet.

elucash:
thank you for PR, please see my reply on  #606

PR #617, opened 2017-05-19
fixed bug for polymorphic classes with primitives inside

asereda-gs:
In some cases mongo repository classes couldn't load data due to the following error:
```
Expected VALUE_STRING but was VALUE_NUMBER_FLOAT
```
GSON lazily loads numbers (to avoid parsing costs) so number may remain string from JsonParserReader perspective.

### DIFF
--- a/gson/src/org/immutables/gson/stream/JsonParserReader.java
+++ b/gson/src/org/immutables/gson/stream/JsonParserReader.java
@@ -27,16 +27,11 @@ import javax.annotation.concurrent.NotThreadSafe;
 import java.io.IOException;
 import java.io.Reader;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.Callable;
 
-import static com.fasterxml.jackson.core.JsonToken.END_ARRAY;
-import static com.fasterxml.jackson.core.JsonToken.END_OBJECT;
-import static com.fasterxml.jackson.core.JsonToken.FIELD_NAME;
-import static com.fasterxml.jackson.core.JsonToken.START_ARRAY;
-import static com.fasterxml.jackson.core.JsonToken.START_OBJECT;
-import static com.fasterxml.jackson.core.JsonToken.VALUE_NULL;
-import static com.fasterxml.jackson.core.JsonToken.VALUE_STRING;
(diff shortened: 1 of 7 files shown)


gson/src/org/immutables/gson/stream/JsonParserReader.java
```diff
@@ -141,7 +144,7 @@ public String nextName() throws IOException {
   public String nextString() throws IOException {
     requirePeek();
     if (!isLenient()) {
-      expect(VALUE_STRING);
+      expectOneOf(VALUE_STRING, VALUE_NUMBER_FLOAT, VALUE_NUMBER_INT);
```
elucash: Major question, maybe we just need to enable lenient mode? Here, `!isLenient()` means "strict mode", in which it would be natural to expect only strings. By accepting `float` and `int` as string means that the mode is no longer "strict" mode. Are there any problems to enable lenient mode?

asereda: I couldn't enable lenient mode using gson builder. Is there an internal setting somewhere ?

The original problem is that [JsonElement TypeAdapter](https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java#L704) does the following optimization: 

```java
      case NUMBER:
        String number = in.nextString();
        return new JsonPrimitive(new LazilyParsedNumber(number));
```

elucash: As far as I see it, current behavior of our subclass breaks things that Gson "expects" to do with all those `LazilyParsedNumber`. I rightfully dislike the change above, but not because a change is not needed, we just doing(or already did) something wrong and we need to revise the check. Let me merge the PR and edit the fragment in question.

asereda-gs: Yes, I agree, this changes expected parser behaviour which is questionable. 
At least there are a couple of tests which should help future fixes.

elucash:
Thank you for the PR, the test suite comes in handy. I've just changed the fragment in question to

```java
if (!peek.isScalarValue()) {
  throw new IllegalStateException("Expected scalar value for string but was " + peek);
}
```
will be released in 2.5.1

PR #621, opened 2017-05-21
Create proper standalone jar for metainf processor

auke-:
The metainf project produces a standalone jar, but the pom file still
declares the shaded dependencies which defeats the purpose of creating a
shaded standalone jar.

By aligning the metainf pom file with the value pom file a proper
standalone jar is created for the metainf processor. This removes the
standalone classifier and creates a standalone jar by default.

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -52,6 +52,14 @@
       <optional>true</optional>
     </dependency>
     <dependency>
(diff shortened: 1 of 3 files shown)


elucash:
Thank you for the PR! I'll take another look before releasing, but seems we're ok.

PR #626, opened 2017-05-25
added atomic operation andReplaceFirst for optimistic locking

asereda-gs:


elucash:
Excellent work! Thank you!

PR #639, opened 2017-06-11
added (eg. readonly) flags to customize repository generator

asereda-gs:
Added two `@Mongo.Repository` annotation attributes:
1. `readonly`. Setting this flag to `true` will remove `delete` / `insert` / `upsert` / `andModifyFirst` ...  write operations from repository. The default value is `false`
2. `index`. Setting this flag to `false` will disable indexer functionality of the repository.


elucash:
Excellent, thank you!

PR #650, opened 2017-07-10
Don't generate duplicate @SuppressWarnings

illicitonion:
If the type suppresses either all or unchecked, it is unnecessary to
also suppress that warning on a method on the type.

The Buck project compiles with ecj with
org.eclipse.jdt.core.compiler.problem.unusedWarningToken=error where
this is flagged as an error.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -1201,7 +1201,7 @@ public interface BuildFinal[type.generics] {
       [/for]
 
 [for bs = type.buildFromTypes]
-[if bs.hasWildcards]
+[if bs.hasWildcards and (not type.suppressesUncheckedWarning)]
   @SuppressWarnings("unchecked")
 [/if]
   private void from(Object object) {
@@ -2557,7 +2557,7 @@ public int ordinal() {
 [/if]
  * @return {@code true} if {@code this} is equal to {@code another} instance
  */
-[if type.generics]@SuppressWarnings("unchecked")[/if]
+[if type.generics and (not type.suppressesUncheckedWarning)]@SuppressWarnings("unchecked")[/if]
 @Override
 public boolean equals([atNullable]Object another) {
   [if type.useReferenceEquality]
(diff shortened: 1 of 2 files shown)


elucash:
@illicitonion  thank you for the PR! It would definitely make sense, but I want to make sure what we're doing. So I initially added the annotation with suppress unchecked, because `"all"` would not cover it. I.e. when you have `@SuppressWarnings("all")` on the top of the top-level class in the file, it would still complain about `"unchecked"`. If this strange behavior was changed in Javac, I just need to make sure we no longer need to care about this case. It may also be that I'm not sure what's happening at all, so if possible, can you please elaborate a bit about compilers affected, (javac,ecj, versions at least approximately, and any observed NDEs with the current approach.

illicitonion:
I put a repro at https://github.com/illicitonion/repro-immutables/tree/unchecked

Basic information:
Compiling with ecj 4.6.1 (also repros with 4.6.2)
Explicitly enabling org.eclipse.jdt.core.compiler.problem.unusedWarningToken=error

Error I see:
```
    [javac] Compiling 1 source file to /Users/dwh/src/github.com/illicitonion/repro-immutables/gen
    [javac] ----------
    [javac] 1. ERROR in /Users/dwh/src/github.com/illicitonion/repro-immutables/gen/example/ImmutableValueWithGenerics.java (at line 63)
    [javac] 	@SuppressWarnings("unchecked")
    [javac] 	                  ^^^^^^^^^^^
    [javac] Unnecessary @SuppressWarnings("unchecked")
    [javac] ----------
    [javac] 1 problem (1 error)
```

If I remove the `@SuppressWarnings("all")` from the top generated class and compile it, I get one warning (a raw type warning for the private ImmutableValueWithGenerics(ImmutableValueWithGenerics original, T optional) constructor)

If I remove the `@SuppressWarnings("unchecked")` from the equals implementation, everything is fine.

I guess the "safe" way of solving this problem would be to generate code which doesn't produce warnings in the first place, and kill the `@SuppressWarnings("all")` on the class... What do you think?

elucash:
ok, let's try it out. I'll merge this and try in build tools/compilers I can reach.

elucash:
@illicitonion  thank you, Daniel!

PR #652, opened 2017-07-13
Nested builders

daicoden:
Feature proposal: Chained builder API.

This PR replicates some of the protobuf builder API.  Given two Builders:

```
@Immutable
@Style(nestedBuilderApi = true, depluralize = true, depluralizeDictionary = {":list"})
public abstract FirstImmutable {
  public abstract SecondImmutable getSingleField();
  public abstract List<SecondImmutable>()  getSecondImmutableList;
}
@Immutable
public abstract SecondImmutable {
  public String getValue();
}
```
You now have this API.

```
ImmutableFirstImmutable.Builder firstBuilder = ImmutableFirstImmutable.builder();
firstBuilder.getSingleFieldBuilder().setValue("hello");
firstBuilder.addSecondImmutableBuilder().setValue("big");
firstBuilder.addSecondImmutableBuilder().setValue("world");
firstBuilder.build();
```

Furthermore, you can use the `@Value.Style.NestedBuilder` to adapt third party builders to this API.  So if you were using immutables with protobufs you can add new proto builders.

I've found for complex graphs, this API is really nice.

I'm opening this to see if there is any interest in the nested (chained?) builder functionality.  If there is, I'd be happy to polish up the pull request, but have a few questions on how things should work. It's at the point where we can use it internally.

If not interested we can close this PR.


daicoden:
Tests aren't going to pass because there's a chicken and egg problem referencing the `@Value.Style.NestedBuilder`, if we go forward with this PR, I can open a separate pull which adds the NestedBuilder (ChainedBuilder?) style API.

elucash:
@daicoden Thank you for the PR! Much appreciated efforts especially in the light of that there were plans for adding external builder support. It will some time for me to review the PR and align it with what I had in mind. Let me get back to it in 1-3 days to leave more detailed feedback. Thank you!

elucash:
I've reviewed the proposed solution and I really like how comprehensive nested builder support and overall implementation approach seems to be in line. Aside from small refinements (which we can do later on), I would like to propose following major changes:

1. Having thought about it some time, I think we can and should reduce our annotation API significantly, up to the point that we should not need `@NestedBuilder` and `@FromMethod` etc annotations. We will lose some flexibility and preciseness, but it will greatly streamline how nested builders are used without losing all the power of the solution. Basically, once nested builder flag is enabled, we should be able to scan attributes and auto-detect nested builder. We would have defaulted naming patterns in our style definition (see how we have an array of patterns to detect `get` accessors), so we would filter enclosed method and class definitions and when matching we would check names, method attributes, parameter and return types to discover the nested builder. If you're new to working with `javax.lang.model` classes or having trouble working with it, let me know can take some efforts to implement it.

2. In order to integrate nested builders from value types (implemented by Immutables processor), `ValueAttribute.attributeValueType` model should be used as it's used when `deepImmutablesDetection` is enabled. Otherwise, auto-detection routines would try to match Builder/builder/build/from methods as from arbitrary objects (see 1. above)

and just to mention, after the solution will be merged, I've planned to add (when Java 8) `builder.attr(Consumer<Attr.Builder>)` method so we can initialize nested builder using a lambda:

```java
buider.attr(b -> b.nested(1).nestid("a"))
```

daicoden:
Thanks @elucash, really like the `builder.attr(b -> b.nested(1).nestid("a")`.

Quick naming question, good to go forward with nested builder? The other name I came up with was chained builder.  Any other names?

>  up to the point that we should not need @NestedBuilder and @FromMethod etc annotations

Good point, I can definitely make it so if you have nestedBuilderApi enabled, it auto-discovers nested builders. The thing that spurred me to create the big API was to support 3rd party builders, a la https://github.com/immutables/immutables/pull/652/files#diff-c9ccf0a941accd8023647277689c9e87R34

I could remove the third party builder support from this PR, and then open a second one later so we could discuss the API more.  Do you think third party nested builder support is something you'd like to have too?

> ValueAttribute.attributeValueType model should be used as it's used when deepImmutablesDetection

Ok perfect, I have another question related to this, so will comment in the code at the relevant location.

value-fixture/src/org/immutables/fixture/builder/FirstPartyImmutableWithDifferentStyle.java
```diff
@@ -0,0 +1,16 @@
...
+import org.immutables.value.Value.Style;
+
+@Immutable
+@Style(
+    builder = "getTheBuilder",
+    build = "doIIT",
+    from = "makeDaCopy",
+    typeBuilder = "Abonabon"
+)
+public abstract class FirstPartyImmutableWithDifferentStyle {
```
daicoden: @elucash - this is the case I'm having trouble with.

Supporting nested builders, when the nested builder style is different than the parent.

value-processor/src/org/immutables/value/processor/meta/NestedBuilderReflection.java
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  @Lazy
+  public ValueType getNestedImmutableValueType() {
+    if (!isImmutable()) {
+      throw new RuntimeException(
+          "You should not be calling getNestedImmutableValueType for non immutable.");
+    }
+
+    ValueType nestedImmutable = new ValueType();
```
daicoden: @elucash this method needs to return the `ValueType` of the nested immutable.

The reason I need the ValueType is to look up the style information.  However, the style information is always the parent style, not the child style.  Is there something different to do here?  Is there a map somewhere with all the different ValueTypes based on a DeclaredType?

I want to make sure the Style follows all the same rules, like if you `@MetaInf` or associate a package with a style, etc.

elucash:
Third party builder support would be fully covered by auto-discovery. The styles should support several (an array of) namings to detect different variations of how builder/build methods are detected, This would work similarly to how we could speficy and detect number of `get` accessor patterns. Looking at the test examples, the third party builder with "generateNewBuilder" "doTheBuild" can be specified as literal or parameterized templates:  "*NewBuilder" , "*Build". I bet that more often than not those methods will follow the convention of some sort. Even if those are completely free-form, pattern could be set to "*" and on the next step we would still match parameter and return types and so can more or less reliably guess various builder and from methods. I can provide more detailed examples if needed.

value-fixture/src/org/immutables/fixture/builder/FirstPartyImmutableWithDifferentStyle.java
elucash: In theory, if style would allow auto-discovery, then we would just detect how particular type's is integrated. I cannot be 100% sure about that, but it seems that we are losing very little if anything if we not using precise configuration (overridable on per-attribute level) for custom builders. The general style would allow number of naming patterns, and auto-discovery would use those patterns along with checking parameter and return types to figure out how particular builder is integrated.

value-processor/src/org/immutables/value/processor/meta/NestedBuilderReflection.java
elucash: Please see `ValueAttribute. initAttributeValueType()`. Basically, we can expand if condition to include `nestedBuilderApi()` (or better `nestedBuilderDetection/Discovery`)

```java
if ((containingType.constitution.style().deepImmutablesDetection()
    || containingType.constitution.style().nestedBuilderDetection())
     && containedTypeElement != null ....
```

once having `attributeValueType` initialized, we can then use to derive any builder models, the attribute value type should have its own style, not necessarily the one used by value type which contains the attribute.

daicoden:
Ah, I like it! I think I understand and can make that work, will give it a shot.

The one case I'm not sure auto discovery will work for is when you turn on factory builders.  I actually am not 100% sure how that feature works, but I'm guessing it make you create a factory `new FooFactory()` instead of `ImmutableFoo.builder()`. So there's nothing that can be discovered from the value object. (I don't have a solution for factories in the current proposal either).

For now I won't worry about factories. If you think factory style builders should work let me know, because I'm missing something.

value-fixture/src/org/immutables/fixture/builder/FirstPartyImmutableWithDifferentStyle.java
daicoden: fixed

daicoden: nice it worked!

daicoden:
Alright, I think there are two pieces of work still to be done to fully use it for my use case.

1) match API with #653 
2) I still need to test a bunch of variants. i.e. no Guava available, combination of third party instance/class methods for various builder functionality. - 
3) My proto classes can be really large, this change makes compiling nested proto builders slow. I was thinking that we could allow developers to register a `NestedBuilderReflection.Strategy` through `@metainf` so you can provide an optimized generator. - open to other ideas.

I'll try and get 1 and 2 done this week.
I'll open a separate pull request for 3.

Anything else that needs to be done?

elucash:
Thank you for the efforts! I think it would be pretty cool functionality. I'm skeptical about extensibility here, we should do our best to optimize it and provide best out-of-the-box experience we can. But given that it's optionally enabled functionality, this should not impact unsuspecting users of the next release. Surely, integrating all together and chase all the details is bunch of work, but overall we're good.

daicoden:
Man, I feel like I was a markov script while cleaning up the template... I think we could do it cleaner, so any pointers welcome. But I through a lot of tests at it, so looking good.

elucash:
so far I very like what I see. Excellent work! Have to look at it some more time, to try it out. Give me at least one day, you will probably also make some refinements in a meantime. Thank you!

value-processor/pom.xml
```diff
@@ -63,7 +63,7 @@
     <dependency>
       <groupId>org.immutables</groupId>
       <artifactId>value</artifactId>
-      <version>2.2.10</version>
+      <version>2.5.6</version>
```
daicoden: This is because I'm using a staged builder, which is after 2.2.10.

value-processor/src/org/immutables/value/processor/meta/AttributeBuilderReflection.java
```diff
@@ -0,0 +1,434 @@
...
+ */
+@Immutable(builder = false)
+@Style(visibility = ImplementationVisibility.PRIVATE)
+public abstract class AttributeBuilderReflection {
+
+  static Map<TypeMirror, AttributeBuilderReflection> analyzedReturnTypes = new HashMap<>();
+
+  //TODO: not 100% cached because returnType can be a list (or map etc.)
+  // Maybe bad to cache? Some safe way to cache without holding onto processing environment?
+  public static AttributeBuilderReflection forValueType(ValueAttribute valueAttribute) {
```
daicoden: the caching here may be dangerous, saw another comment that ValueAttribute has a handle to the environment which can cause memory leaks. Better way to cache?

elucash:
best way to cache the discovery information is to have string key for
fully-qualified name (raw form) and make sure descriptor `
AttributeBuilderReflection` do not capture any `javax. lang.model*` stuff,
only strings rendered. There's ValueType.containedTypeElement field which
contains either attribute type, or element type in case of collections,
using it will also defend against lot of cases such as generic Type
variable instead of a type, so `containedTypeElement ` would just be null
for all those cases. From there you can
`containedTypeElement.getQualifiedName().`
to use as a key.

daicoden:
I had to make a pretty big change to the Third Party API support inner classes, sorry if you already have been taking a look.

However, I think this PR now supports all the use cases you mentioned! (only supports getBuilder and addBuilder, but the rest is just new features in the template)

elucash:
no worries, I will be getting to it only tomorrow )   Glad to see the progress. Thank you!

value-processor/src/org/immutables/value/processor/meta/AttributeBuilderReflection.java
```diff
@@ -0,0 +1,463 @@
...
+   * Assumes that all third party classes are available on the class path at processing time.
+   *
+   * If this is not the case, we may need to use a processing method that allows deferring of
+   * compilation. Example, moving from auto-value to immutables... though maybe you would just
+   * implement a new strategy for that use case...
+   */
+  @Immutable(builder = false)
+  abstract static class ThirdPartyStaticBuilderStrategy implements Strategy {
+
+    private ExecutableElement buildMethod = null;
```
daicoden: grumble... so this isn't an immutable anymore... not sure the best way to clean this up unfortunately. It's a lot cleaner this way than before when I had a bunch of lazy optionals.

daicoden:
Friendly ping on this :)

cc @ijustdontgitit

elucash:
sorry, was a little bit overwhelmed with other stuff. Seems like we're reached a good state. I may make some minor refinements when I'll get closer. I'll merge it for now. The soonest release would be, probably, over the weekend.

daicoden:
Weee awesome! Thanks so much! Was a very pleasant experience working with you @elucash ! Let me know if you would like help adding the `builder((lambda) -> awesome)` proposal

PR #653, opened 2017-07-17
Nested builder detection API.

daicoden:
Opening this @elucash to make sure we're on the same page about the API.

If this looks good to you, it would be helpful to merge this, do a release, so I can update value-processor/pom.xml.

### DIFF
--- a/value/src/org/immutables/value/Value.java
+++ b/value/src/org/immutables/value/Value.java
@@ -1149,6 +1149,94 @@ public @interface Value {
     String redactedMask() default "";
 
     /**
+     * When enabled: immutable attributes with discoverable builders receive the additional
+     * builder API:
+     *
+     * For single children:
+     * {@code BuilderT *Builder()}
+     * {@code ParentT *Builder(BuilderT builder)}
+     *
+     * For a collection of children:
+     * {@code BuilderT add*Builder()}
+     * {@code ParentT addAll*Builder(Iterable<BuilderT> builderCollection)}
+     * {@code ParentT addAll*Builder(BuilderT... builderArgs)}
+     * {@code List<BuilderT> *Builders()}
+     *
+     * In strict mode, you may only set the builder via {@code *Builder(BuilderT builder)} once,
+     * but you may call {@code *Builder()} multiple times, in which the same builder is returned.
+     * If the nested immutable is also strict, then you will only be able to set properties on
+     * the child builder once.
+     *
+     * To discover builders on value attributes the value methods are scanned for method names
+     * matching a patterns specified in {@link #attributeBuilder()}.
+     *
+     * This style parameter is experimental and may change in future.
+     * @return true to enable the feature.
+     */
+    boolean attributBuilderDetection() default false;
+
+    /**
+     * Pattern for detecting builders.
+     *
+     * {@link #attributeBuilder()} applies to both
+     * static and instance methods. In the case a builder is only discoverable through a
+     * value instance method, the builder class must have a public no-arg static construction
+     * method. To use a no-arg public constructor, a special token "new" should be specified.
+     *
+     * <em>This is detection pattern, not formatting pattern. It defines how to recognize a nested builder.</em>
+     * Only applies if {@link #attributBuilderDetection()} is {@code true}.
+     *
+     * @return naming template
+     */
+    String[] attributeBuilder() default {"*Builder", "builder"};
+
+    /**
+     * Naming template for retrieving a nested builder.
+     *
+     * Only applies if {@link #attributBuilderDetection()} is {@code true}.
+     *
+     * @return naming template.
+     */
+    String getBuilder() default "*Builder";
+
+    /**
+     * Naming template for setting a nested builder.
+     * This may be called only once in strict mode.
+     *
+     * Only applies if {@link #attributBuilderDetection()} is {@code true}.
+     *
+     * @return naming template.
+     */
+    String setBuilder() default "*Builder";
+
+    /**
+     * Naming template for adding a new builder instance to a collection.
+     *
+     * Only applies if {@link #attributBuilderDetection()} is {@code true}.
+     *
+     * @return naming template.
+     */
+    String addBuilder() default "add*Builder";
+
+    /**
+     * Naming template for adding a collection of builders.
+     *
+     * Only applies if {@link #attributBuilderDetection()} is {@code true}.
+     *
+     * @return naming template.
+     */
+    String addAllBuilder() default "addAll*Builder";
+
+    /**
+     * Naming template for retrieving an immutable list of builders.
+     *
+     * Only applies if {@link #attributBuilderDetection()} is {@code true}.
+     *
+     * @return naming template.
+     */
+    String getBuilderList() default "*Builders";
+
+    /**
      * If implementation visibility is more restrictive than visibility of abstract value type, then
      * implementation type will not be exposed as a return type of {@code build()} or {@code of()}
      * constructon methods. Builder visibility will follow.


value/src/org/immutables/value/Value.java
```diff
@@ -1149,6 +1149,93 @@
...
+     * static and instance methods. In the case a builder is only discoverable through a
+     * value instance method, the builder class must have a public no arg constructor,
+     * or a no-arg static method constructor.  The static method will be preferred.
+     *
+     * <em>This is detection pattern, not formatting pattern. It defines how to recognize a nested builder.</em>
+     * Only applies if {@link #nestedBuilderDetection()} is {@code true}.
+     *
+     * @return naming template
+     */
+    String[] nestedBuilderDetectionPattern() default {"*Builder", "*builder"};
```
elucash: `"*builder"` seems unnecessary as it will match `adcbuilder` or `xyzbuilder`, but not `builder`.
better default is `{"*Builder", "builder", "new"}` where `"new"` is a special token indicating that plain constructor can be used. This would match constructors and factory methods like `builder, newBuilder, createBuilder, instanceBuilder` etc

value/src/org/immutables/value/Value.java
```diff
@@ -1149,6 +1149,93 @@
...
+     * times, in which the same builder is returned. If the nested immutable is also strict,
+     * then you will only be able to set properties on the child builder once.
+     *
+     * To discover builders on value attributes the value methods are scanned for method names
+     * matching a patterns specified in {@link #nestedBuilderDetectionPattern()}.
+     *
+     * This style parameter is experimental and may change in future.
+     * @return true to enable the feature.
+     */
+    boolean nestedBuilderDetection() default false;
```
elucash: This popped up from the top of my head just now: we could also apply "attributeBuilder" term for the functionality. As I also like `nestedBuilder` as it is here, I cannot decide one over another, so perhaps it's better to stick to "nested"

value/src/org/immutables/value/Value.java
```diff
@@ -1149,6 +1149,93 @@
...
+    String[] nestedBuilderDetectionPattern() default {"*Builder", "*builder"};
+
+    /**
+     * Naming template for retrieving a nested builder.
+     *
+     * Only applies if {@link #nestedBuilderDetection()} is {@code true}.
+     *
+     * @return naming template.
+     */
+    String getBuilder() default "get*Builder";
```
elucash: I would prefer `default *Builder"` here: by default, attribute initializers do not use `set` prefixes so this would be ok

value/src/org/immutables/value/Value.java
```diff
@@ -1149,6 +1149,93 @@
...
+
+    /**
+     * Naming template for setting a nested builder.
+     * This may be called only once in strict mode.
+     *
+     * Only applies if {@link #nestedBuilderDetection()} is {@code true}.
+     *
+     * @return naming template.
+     */
+    String setBuilder() default "set*Builder";
```
elucash: I would prefer `default *Builder"` here: by default, for attribute initializers in builder we do not use `set` prefixes, I believe there should be no conflict with `getBuilder` pattern because of the parameter type differs (properly overloaded method)

value/src/org/immutables/value/Value.java
daicoden: ah, I thought * was like a regex and meant 0 or more. Will change.

value/src/org/immutables/value/Value.java
```diff
@@ -1149,6 +1149,93 @@
...
+    String addAllBuilder() default "addAll*Builder";
+
+    /**
+     * Naming template for retrieving an immutable list of builders.
+     *
+     * Only applies if {@link #nestedBuilderDetection()} is {@code true}.
+     *
+     * @return naming template.
+     */
+    String getBuilderList() default "get*BuilderList";
```
elucash: `String getBuilders() default "*Builders";` is preferred. We almost always use plurals instead of `List` for API like this, although it's just a preference

value/src/org/immutables/value/Value.java
daicoden: I like attribute builder too... Nested doesn't quite define what this is for me. would nestedAttributeBuilder be too much?

value/src/org/immutables/value/Value.java
elucash: yeah, that is for the most part an unfortunate coincidence that star in "*builder" is not working as zero or more. Actually, it captures zero chars for `"builder"` input and the return value of empty string is considered as detection failed. Anyway, people almost always concatenate with camel-case `Builder` suffix, so `*builder` seems unpractical

elucash:
Thank you for the API PR. API seems to be pretty neat now. Please, also consider if we need to add following detection patterns:
* Patterns for nested builder type detection `default {"Builder"}`
* Patterns for build method detection `default {"build", "build*", "create"}`

If I would name it, here's what my randomizer device tells me:
* `nestedBuilderCreatePattern` (possibly renamed from `nestedBuilderDetectionPattern`) (variants `Create`, `Factory`, `Construct`, `New` suffixes
* `nestedBuilderTypePattern` (in other occurences i've placed `type` prefix in front, so possible variant is ``)
* `nestedBuilderBuildPattern`

variant: remove `Pattern` from these names, I don't see we ever used `Pattern` or `Naming` suffix, so we can omit it here. There's almost nothing intuitive about those names so all hopes are for consistency and Javadoc to explain it. I'm thinking as I write, so possibly most consitent would be `newNestedBuilder`, `typeNestedBuilder`, `buildNestedBuilder`. I dunno

value/src/org/immutables/value/Value.java
elucash: `nestedAttributeBuilder` will explain it well. We can apply it everywhere, longer but quite descriptive

daicoden: What about just `nestedAttribute` ? or `nestedAttributeApi`

elucash: seems "builder" should be there, as it's all about builders

elucash: `attributeBuilder` seems like precise, we provide builders for attributes. "Nested" is something more overloaded.

daicoden:
> Please, also consider if we need to add following detection patterns.

I updated #652, take a look but I don't think we need the build detection pattern.  Once you find a pattern which matches the `nestedBuilderDetectionPattern`, you can then find the build method by a no-arg method that returns the same type (or a type that is an ancestor thereof).

Regarding the nested type detection. I'm thinking we could also get away without a new pattern because we would only investigate inner classes if the "new" token is in the `nestedBuilderDetectionPattern`. For inner classes which have builder methods on the value type, it's already handled.

I still need to update this PR with the other changes.

elucash:
yes, I agree in general. well, if needed, we can have implicit `{"*Builder", "Builder"}` matching for nested builder types if "new" enabled. (star can even be substituted with simple name of the attribute type, so for type `Foo` we can match `"FooBuilder"` and `"Builder"` ). I wonder if it's necessary at all, but just in case.

daicoden:
Take a look. Leaving the Builder discovery out for now, but will use implicit `{"*Builder", "Builder"}` if needed.

elucash:
Thumbs up! merged

daicoden:
Woot! Want me to open a PR with a version bump? I guess I should have done that here.

elucash:
I will do this while releasing anyway, we can skip it for now

daicoden:
roger, when you get a chance if you could release 2.5.7, I can get #652 green. Reason it's not green is because value-processor/pom.xml relies on 2.2.10, and can't see the new API.

elucash:
value-processor/pom.xml relies on older 2.2.10 to compile itself, however it's fully decoupled from API we have in `value` module. We use special mirror generation to achieve that, we only need that API match the stubs in `ValueMirrors` and `Style` attributes is propagated via `StyleInfo` object. See how it's done for other style properties, let me know if it will not work for some reason, but I believe you can proceed freely. Maybe we're facing unrelated issues.

daicoden:
I was wondering what the mirror was! I think the problem was I was using the `@NestedBuilderApi` annotation, and didn't create a mirror. Makes sense now, I think I can get it working.

PR #660, opened 2017-07-25
mongo support for subtyped comparable like LocalDateTime

michaelmosmann:
If you use LocalDateTime as a property: 

`
@Repository
public interface Foo {
  LocalDateTime bar();
}
`

.. the generated FooRepository code will not compile.


elucash:
Thank you! It seems straightforward!

Issue #668, opened 2017-08-07
GSON @SerializedName must be passed on to the value, not getter.


alexanderkjeldaas:
When using GSON on a class like so:

```
@Value.Style(passAnnotations = {SerializedName.class})
public abstract class Foo {

    @SerializedName("@id")
    public abstract String id();
}
```

immutables will place this annotation on the getter:

```
  ...
  private final String id;
  ...
  /**
   * @return The value of the {@code id} attribute
   */
  @SerializedName("@id")
  @Override
  public String id() {
    InitShim shim = this.initShim;
    return shim != null
        ? shim.id()
        : this.id;
  }
```

However, GSON wants it on the field, like this:

```
  ...
  @SerializedName("@id")
  private final String id;
  ...
  /**
   * @return The value of the {@code id} attribute
   */
  @Override
  public String id() {
    InitShim shim = this.initShim;
    return shim != null
        ? shim.id()
        : this.id;
  }
```

Using the generated type adapters is one option, but it might be untenable for mobile applications where the size of the generated code is too high (it's around 15kB per class which can easily add up to several MBs of additional code).

PR #669, opened 2017-08-07
Apply SerializedName to the immutable member fields?

alexanderkjeldaas:
Untested possible fix for #668

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -2227,11 +2227,9 @@ this.[n] = [valueFromValue v][invokeSuper v].[v.names.get]()[/valueFromValue];
   [rr.declareFields v]
   [else]
   [jsonIgnore type]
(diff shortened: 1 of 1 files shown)


elucash:
@alexanderkjeldaas Hi. sorry, I actually never got to this before. My sincere apologies for not giving any feedback in time. Part of it is that I didn't understand the context of this PR. `@SerializedName` is handled by Gson type adapters generator and expects it on abstract attribute accessor. If we put this on the field, it can only be consumed by standard/default reflective type adapter in Gson, and basically, it's not an intended usage. On the other hand, I'm in process of development of general-purpose annotation passing functionality/configuration which will allow propagating arbitrary annotations inside generated classes, i.e. on generated fields, constructor and builder parameters, etc.

elucash:
closing this, thank you. There's now general functionality to push/inject annotation to the generated elements (fields, accessors, initializers) elements http://immutables.github.io/immutable.html#annotation-injection

PR #677, opened 2017-08-23
Fix nondeterminism in code generation of nested values

jkeljo:
Because `HashSet` iteration order can vary from run to run, the order in which the processor
was generating nested value classes could change from run to run. This caused problems with
build systems like Buck that assume deterministic output so that artifacts can be cached.

Switching to a `LinkedHashSet` takes care of the issue.


elucash:
Thank you! in general I tried to use predictable order sets and maps, apparently, the are some places like this, thnx!

jkeljo:
Cool! Thanks for the speedy merge.

PR #688, opened 2017-09-14
minimal effort migration to Mongo driver v3

asereda-gs:
Some changes to be able to use new mongo v3 driver (currently 3.4.3). There is still refactoring work to be done (internally) to fully leverage new v3 functionalities decoders / readers etc. This is backwards-compatible change.

Probably `bson4jackson` dependency can be dropped in favor of new [BsonReader](http://api.mongodb.com/java/3.3/?org/bson/BsonReader.html) / [BsonWriter](http://api.mongodb.com/java/3.3/?org/bson/BsonWriter.html) (stream friendly?) API assuming performance results are satisfactory. 

Also since [v3.x](http://mongodb.github.io/mongo-java-driver/3.5/whats-new/) they have added [Codec serialization / deserialization](http://mongodb.github.io/mongo-java-driver/3.0/bson/codecs/) perhaps, in this case, mongo-Gson dependency can be also deprecated ? It is a more complicated change.


michaelmosmann:
did some smaller changes ( https://github.com/immutables/immutables/pull/691 ) ... but i am not sure, if this pull request is the better way to do this..

anyway..

i am very interested in any solution which brings mongo driver 3.5.x to immutables.org

elucash:
Looks gorgeous, thank you!

PR #691, opened 2017-09-20
change mongodb driver to 3.5.0

michaelmosmann:


asereda-gs:
Hi Michael,

I'm not sure this will work with real Mongo. Can you run the test with the following flag (inside `immutables/mongo` folder) :

```
mvn test -DargLine="-Dmongo=mongodb://localhost"
```
Obviously you need a running instance.

PS. Any reason why not use existing PR #688 for mongo v3 driver migration ?

michaelmosmann:
thanks.. you are right.. it does not work with a real mongodb...

Issue #392, opened 2016-07-08
Lint warnings for generic builder methods taking variadic arguments


jfancher:
Example:

````java
@Value.Immutable
interface Bug<E> {
  Set<E> getItems();
}
````

This does not compile cleanly with `javac -Xlint` because `Varargs method could cause heap pollution from non-reifiable varargs parameter elements` on each of the methods that takes a `E...` and passes it as an argument to a method that takes a `E[]`. This warning is not affected by `@SuppressWarnings("all")` nor by `@SafeVarargs`.

A solution is to add `@SuppressWarnings("varargs")`. This may not be great, because it's very javac-specific. But the only other solution I've found is to change the implementation of the method to iterate the array instead of passing it around.

Alternatively (or additionally), I don't find the varargs overloads to be all that useful... maybe the simplest fix would be to have a style option to turn them off?

elucash:
Agree, will try to do something about it. It came on radar once, but did not received enough attention then. Thank you for reporting this!

elucash:
Getting back to this, there are no much issue left as `@SafeVarargs` is generated on the `(E...)` initializer, with the fix to #393, there's possible to specify suppression for `"varargs"`.
The encoding functionality, currently in final stage of development, will allow to override/specify a custom attribute handling, so if some overloads are just painfully unnecessary, then it will be possible to provide own set of methods instead (see #363). Thank you!

PR #694, opened 2017-09-25
Extend safe varargs annotation usage

Stephan202:
- Add `@SafeVarargs` in more relevant places.
- Where `@SafeVarargs` is used, also add `@SuppressWarnings("varargs")`.

The suppression is not necessary in all cases, but avoids the issue described
in #392, making Immutables more friendly towards users compiling with
`-Xlint:all` or `-Xlint:varargs`.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -574,9 +574,11 @@ private Object writeReplace() {
   [for v in type.settableAttributes]
     [if v.arrayType]
 @Override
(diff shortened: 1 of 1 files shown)


Stephan202:
(A workaround is to apply `@SuppressWarnings("varargs")` to the `@Value.Immutable`-annotated type, but that will just confuse the casual reader unless a comment is added. Since the warning pertains to an Immutables implementation detail, I think it makes sense to generate the suppressions on the relevant methods.)

elucash:
Thank you!

PR #698, opened 2017-10-06
Nested immutable with primitive

daicoden:


daicoden:
I found this bug while using attributeBuilders.

We look for `valueObject.toBuilder()` as part of the detection. Getters also look like the toBuilder() method - no arg instance method.  This causes a cast error later.

Since a primitive can never be a builder, we can return false immediately.

elucash:
Thank you!

PR #706, opened 2017-10-20
Support Error Prone's `@Immutable` annotation

Stephan202:
Introduces support for [`@com.google.errorprone.annotations.Immutable`](http://errorprone.info/bugpattern/Immutable) by selectively suppressing known-safe mutable implementation details in the generated code.

### DIFF
--- a/value-fixture/pom.xml
+++ b/value-fixture/pom.xml
@@ -30,6 +30,10 @@
     Module that contains all tests for the code generation capability
   </description>
 
+  <properties>
+    <errorprone.version>2.0.19</errorprone.version>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>org.immutables</groupId>
@@ -146,6 +150,11 @@
       <artifactId>hibernate-validator</artifactId>
       <version>5.4.0.Final</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.errorprone</groupId>
+      <artifactId>error_prone_annotations</artifactId>
+      <version>${errorprone.version}</version>
+    </dependency>
   </dependencies>
   <repositories>
     <repository>
@@ -178,7 +187,7 @@
           <dependency>
             <groupId>com.google.errorprone</groupId>
             <artifactId>error_prone_core</artifactId>
-            <version>2.0.19</version>
+            <version>${errorprone.version}</version>
           </dependency>
         </dependencies>
       </plugin>
--- a/value-fixture/src/org/immutables/fixture/AllMandatoryParams.java
+++ b/value-fixture/src/org/immutables/fixture/AllMandatoryParams.java
@@ -28,6 +28,7 @@ public interface AllMandatoryParams {
     return "C";
   }
 
+  @SuppressWarnings("CheckReturnValue")
   static void use() {
     ImmutableAllMandatoryParams.of(1, true).withC("ABC");
   }
--- a/value-fixture/src/org/immutables/fixture/CancelParam.java
+++ b/value-fixture/src/org/immutables/fixture/CancelParam.java
@@ -33,6 +33,7 @@ public interface CancelParam extends Param {
   @Value.Parameter(false)
   List<Integer> aux();
 
+  @SuppressWarnings("CheckReturnValue")
   default void use() {
     ImmutableCancelParam.of(1).withAux(1, 2, 3);
   }
--- a/value-fixture/src/org/immutables/fixture/DeprecateType.java
+++ b/value-fixture/src/org/immutables/fixture/DeprecateType.java
@@ -21,6 +21,7 @@ import org.immutables.value.Value;
 @Deprecated
 interface DeprecateType {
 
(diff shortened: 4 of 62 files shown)


value-processor/src/org/immutables/value/processor/meta/ValueType.java
```diff
@@ -428,23 +428,34 @@ public CaseStructure getCases() {
...
+    return customAnnotations;
   }
 
   private ImportsTypeStringResolver newTypeStringResolver() {
     @Nullable DeclaringType type = constitution.protoclass().declaringType().orNull();
     return new ImportsTypeStringResolver(type, type);
   }
 
+  public boolean isErrorProneImmutable() {
+    return passedAnnotations().contains("@com.google.errorprone.annotations.Immutable");
```
Stephan202: This `#contains` check is the reason for the `CharSequence` -> `String` changes. In practice the `CharSequence` instances were `StringBuilder`s, and those don't implement `#equals`. If you'd like me to implement this some other way, just let me know :)

elucash: Thank you, overall, I'm ok with the idea. (Sorry for not getting to this earlier)
Looking at this, just to mention, `CharSequence/StringBuilder` was to try to avoid tons of strings copying, but actual effect/impact have not investigated. As for me it seems that it's a little strange to piggy-back on `passedAnnotations()` accessor as it's complicating things even more (fully realizing that a lot of things are already complicated ;) ). Seems that dedicated support for detecting `com.google.errorprone.annotations.Immutable` and copying it over would be little bit more straightforward. You can just hook into `ValueType` and detect it's presence on `element` via lazy getter, or/and for example, ask it's present in `constitution.style().passAnnotationsNames()` (but may not look into this property at all), or auto-detect on the classpath and apply if necessary (using accessor to private `Environment.findElement`). I'm a bit confused if we would auto-add this to implementation or add it only if it's placed on abstract value type, but it can work without changing CharSequences to Strings (not because it was a good idea to use CharSequences, but just because it's unnecessary twist here, IMHO)

Stephan202: Thanks for the feedback! I'll have to open the code in my IDE to cross-reference your suggestions; hope to get back to this later this week :)

Stephan202: This PR slipped my attention; it's been very busy at work recently. Commit google/error-prone@e7daed7e102c341230905cc07872b778250c4193 was mirrored out yesterday; that makes this PR quite a bit more pressing... so will try to find time to finalize it.

vorburger:
@elucash ping to bump.. did you want to review and merge this one some time?

In the mean time, I'll put Immutables generated code on error-prone's -XepExcludedPaths ...

Stephan202:
@vorburger I think the ball is in my court the update the PR as @elucash suggested. I've just been completely swamped at work for a few weeks now. I'll try my best to wrap this up. (But cannot promise a clear time line. Hopefully before the end of the week.)

vorburger:
@Stephan202 no stress! Thank You for contributing to FLOSS. FYI something that may interest you as well: just after commenting here yesterday, I found https://github.com/google/error-prone/issues/821 which says an error-prone release with support for -XepExcludedPaths is apparently imminent - meaning  one can then work-around code generators such as immutables not conforming to error-prone by putting their output on error-prone's excluded paths.. That will make issues like this one a lot less blocking than they currently are anyway.

Stephan202:
@elucash, finally got back to this PR. I propose to simplify the PR as follows:
* Revert the `value-processor/**/*.java` changes as you requested.
* Apply the suppression whenever the Error Prone annotation is on the classpath, i.e. using
    
    ```
    [if classpath.available 'com.google.errorprone.annotations.Immutable']@SuppressWarnings("Immutable")[/if]
    ```

This seems by far the simplest solution; not sure why I didn't just go for this in the first place. WDYT?

NB: I rebased the branch on `master` and the build now fails because this PR adds `error_prone_annotations` on the `value-fixture` classpath, while #719 raises `CheckReturnValue` violations as errors. Before I proceed with this PR I'll prepare a separate PR to deal with this. (Fixing it will amount to adding `@SuppressWarnings("CheckReturnValue")` in a bunch of places, which is very unfortunate, but on the plus side it appears there is also a genuine bug to be fixed there.)

elucash:
Totally makes sense, thank you for heads-up

Stephan202:
Thanks for quickly validating the alternative approach. I filed #728 for the build failure; I hope you can live with all those `@SuppressWarnings("CheckReturnValue")`. (Alternative suggestions welcome!)

Stephan202:
I have rebased the branch on top of the fix for #728 and squashed the suggested changes, cause in this particular case that seems to make reviewing actually simpler. Locally I created a tag in which the changes are applied as separate commits on top of the original PR submission, so if you prefer that I'll force-push the branch once more to re-split the work.

PR #718, opened 2017-11-05
Handle Error Prone's `MethodCanBeStatic` check

### LINKED ISSUE #717: static checkNotIsSet method to satisfy error-prone's MethodCanBeStatic bug pattern
Would you consider marking the generated static checkNotIsSet() method as static?

    private void checkNotIsSet(boolean isSet, String name) {
      if (isSet) throw new IllegalStateException("Builder of ThreadFactoryProvider is strict, attribute is already set: ".concat(name));
    }

this would have no impact, and would satisfy [error-prone's MethodCanBeStatic bug pattern](http://errorprone.info/bugpattern/MethodCanBeStatic).

PS: Due to https://github.com/google/error-prone/issues/808, one cannot currently enforce  MethodCanBeStatic in a project using Immutables; that's error-prone's problem though, not Immutables' - just FYI.

Stephan202:
Where feasible, mark generated methods `static`. Where infeasible, add a suppression. The Gson code's complexity is such that I had to add an unconditional suppression at the top of the generated class.

@elucash, this change requires quite careful review, unfortunately. The compiler caught most issues, but if a fixture with a certain combination of features is absent, I might have missed an edge case.

Aside from appeasing Error Prone users, I think this change is worthwhile because use of `static` may allow micro-optimizations by the JVM, and we do want our Immutables to have efficient implementations. That said, whether the extra complexity introduced here is worth it is up for debate. WDYT?

Fixes #717. CC @vorburger.

### DIFF
--- a/value-fixture/pom.xml
+++ b/value-fixture/pom.xml
@@ -166,6 +166,7 @@
           <target>1.8</target>
           <compilerArgs>
             <arg>-Xep:CheckReturnValue:WARN</arg>
+            <arg>-Xep:MethodCanBeStatic:ERROR</arg>
           </compilerArgs>
         </configuration>
         <dependencies>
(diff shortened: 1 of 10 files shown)


elucash:
Thank you! need some more time to review, though

vorburger:
@elucash ping to bump.. did you want to review and merge this one some time?

In the mean time, I'll put Immutables generated code on error-prone's -XepExcludedPaths ...

elucash:
Thank you, merging this, may need to recheck couple of places, but overall we're ok here

PR #719, opened 2017-11-05
Handle `CheckReturnValue` violations as errors

Stephan202:
Rationale:
- Prevents regressions w.r.t. this feature.
- Proves that the issue reported in https://github.com/google/error-prone/issues/647#issuecomment-341664913 is either already fixed or not covered by current tests.

CC @bmhm.


elucash:
Let's give it a try! Thank you!

PR #721, opened 2017-11-06
Make checkNotIsSet() static for error-prone's MethodCanBeStatic

### LINKED ISSUE #717: static checkNotIsSet method to satisfy error-prone's MethodCanBeStatic bug pattern

vorburger:
see http://errorprone.info/bugpattern/MethodCanBeStatic

Fixes #717 https://github.com/immutables/immutables/issues/717

Signed-off-by: Michael Vorburger <vorburger@redhat.com>

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -1981,7 +1981,7 @@ checkNotIsSet([v.names.isSet](), "[v.names.raw]");
   [/for]
   [if type.useStrictBuilder and (nondefaultsPositions.longs or positions.longs)]
 
-  private void checkNotIsSet(boolean isSet, String name) {
+  private static void checkNotIsSet(boolean isSet, String name) {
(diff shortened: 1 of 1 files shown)


vorburger:
I had actually not noticed https://github.com/immutables/immutables/pull/718 .. thanks @Stephan202 if your goes in first then I'll just Abandon this - otherwise @elucash if there's more discussion needed on yours then perhaps this can already go in just to fix that one problem before the better wider solution makes it in?

elucash:
thank you! that one is easy

PR #728, opened 2017-12-04
Generate test fixtures with Error Prone annotations

Stephan202:
And resolve the issues thus identified:
- Certain generated `@Value.Modifiable` methods are `void`; those must not be
  annotated `@CanIgnoreReturnValue`.
- Encoding builders may return `this`, therefore `@CanIgnoreReturnValue`
  applies.

The remainder of this change consists of the introduction of strategically
placed `@SuppressWarnings("CheckReturnValue")` annotations; these are necessary
due to the nature of the tests.

### DIFF
--- a/value-fixture/pom.xml
+++ b/value-fixture/pom.xml
@@ -30,6 +30,10 @@
     Module that contains all tests for the code generation capability
   </description>
 
+  <properties>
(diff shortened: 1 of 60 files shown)


elucash:
Thank you! merging, need to dive in to see what's happening with those suppressions

PR #735, opened 2017-12-15
close MongoCursor after fetch

asereda-gs:
use `try / catch` block when fetching results from MongoCollection. Otherwise resources are not properly closed.


elucash:
Nice PR! Thank you!

PR #743, opened 2018-01-03
added forced failure if there is no DuplicateKeyException for unit tests

asereda-gs:
Added ```fail()``` where it should have been enforced


elucash:
thank you!

PR #746, opened 2018-01-25
added more tests for numbers

asereda-gs:
Explicit `NumbersTest` to validate number wrappers like `BigDecimal`, `BigInteger`, `AtomicLong` etc.
Also added a separate statement for `Byte` (inside `BsonWriter`) which will persist Byte as int32 (in mongo) instead of [default double](https://github.com/asereda-gs/immutables/blob/449d63b4f9c71e430a699ec75cbd093c383c6cb5/mongo/src/org/immutables/mongo/repository/internal/BsonWriter.java#L190).


elucash:
thank you!

PR #773, opened 2018-05-14
Avoid raw types in some generated code

Stephan202:
This removes 3 out of 18 `-Xlint:rawtypes` warnings in `value-fixtures`. Not a lot, but we happen to have code that triggers this particular case.


elucash:
Merged, thank you. Sorry, I somehow missed this for 2.6.0 release, but this will be in 2.6.1

Stephan202:
No worries :). And good news: preliminary tests of the 2.5.6 -> 2.6.0 upgrade in our code base turn up no regressions and confirm that #694 works as expected.

PR #806, opened 2018-07-24
Fix Javadoc generated for @Value.Lazy

simon04:
Version 2.6.3 generates this Javadoc:

```
...
@return A lazily initialized value of the {@code l.name} attribute
...
```


elucash:
Thank you for the PR!
(sorry for the late response)

Issue #813, opened 2018-08-16
Pattern matching criteria for Optional<String> fields


coreyjboyle:
Hi

When using Immutables generated Repositories, only string fields have pattern matching criteria methods generated (*StartsWith, *Matches, *NotMatches).  We would like to have these methods generated even for Optional<String> fields.  Is this something that makes sense to add to Repositories.generator, or is there a different/better way?  I'm happy to contribute if this is the recommended approach.

Thank you
Corey

coreyjboyle:
Went ahead and created a PR, #814, in case this is the recommended approach.

Thanks!

elucash:
This is straightforward piece of functionality! Merged PR! Thank you!

elucash:
released in 2.7.1 . Thank you!

PR #814, opened 2018-08-16
Add pattern criteria for Optional<String> attributes

coreyjboyle:
Add pattern matching criteria methods (*StartsWith, *Matches, *NotMatches) for Optional<String> attributes.

#813


elucash:
Great job! I guess this is straightforward to allow string specific queries on `Optional<String>`. Thank you for PR!

Issue #792, opened 2018-06-12
At least Builder class is not annotated by `org.immutables.value.Generated`


Godin:
Hi,

I'm one of JaCoCo maintainers, during work on resolution of jacoco/jacoco#656 we discovered that while #756 was marked as resolved, it seems that resolution is incomplete - not all generated classes receive annotation `org.immutables.value.Generated`, in particular Builder classes, maybe others, and so JaCoCo can't filter out such.

This can be observed with version 2.6.1 on a basic example ([example.zip](https://github.com/immutables/immutables/files/2096081/example.zip)) :
for `src/main/java/Example.java`
```java
@org.immutables.value.Value.Immutable
public abstract class Example {
  public abstract int foo();
}
```

generated `target/generated-sources/annotations/ImmutableExample.java` has annotation on a top level class, but not on a `Builder`:
```
$ cat target/generated-sources/annotations/ImmutableExample.java | grep "org.immutables.value.Generated" -A 1
@org.immutables.value.Generated(from = "Example", generator = "Immutables")
public final class ImmutableExample extends Example {
```

And thus it is missing in class file too:
```
$ javap -v -p target/classes/ImmutableExample | grep Generated | wc -l
1

$ javap -v -p target/classes/ImmutableExample\$Builder | grep Generated | wc -l
0
```

I guess that something is missing near https://github.com/immutables/immutables/blob/86cb65b89f9f77ff124d0398e6551a8736f77366/value-processor/src/org/immutables/value/processor/Immutables.generator#L1073
Or maybe I'm doing something wrong?

elucash:
Yes, I think I've overlooked that all this is happening on the class level (as opposed to source-level), so nested builder is just a separated class (with some attributes linking it to the outer and name mangling, but still a standalone class). Should not be a problem to add an annotation to generated inner classes.

ylexus:
Community is holding the breath waiting for this to be implemented. So much ugly and fragile Jacoco exclusion configuration would be avoided if we had this issue addressed. In our case we even hit a jacoco command line limit as we needed to explicitly exclude a lot of various immutable generated classes in various packages.

elucash:
sorry for any delays, I'll release fixed version over the weekend (97% probability)

elucash:
the change is there, but I still messing with some other changes for the release

PR #815, opened 2018-08-18
Make builder fields final when possible

Stephan202:
This resolves some instances of Error Prone's [`FieldCanBeFinal`](http://errorprone.info/bugpattern/FieldCanBeFinal) bug pattern. I started working on this before I realized that I should just wait for #792 to be resolved, as the problem will then "go away" in combination with `-XepDisableWarningsInGeneratedCode`. Opening this PR in case you're interested in this contribution regardless.


elucash:
The functionality seems to be innocent (given strictBuilder and toNull handled), thank you for PR! Excellent work!

PR #817, opened 2018-08-22
Allow external CodecRegistry for serialization and querying in mongo

asereda-gs:
Note: This PR is to start a discussion don't merge it yet.

[ ] Backwards compatibility with Gson adapter (eg. `Date` / `ObjectId` / `TimeInstant` etc.)
[ ] Support for Map kinds (`secondaryAdapter`)

This change will allow users to use other serialization frameworks than Gson (eg. jackson or native
 BSON pojo encoders etc.). Backwards compatibility with Gson is preserved.

1. Moved bson/gson adapters to a separate package bson4gson which can be developed independently.
2. Changed RepositorySetup.Builder to accept CodecRegistry (if not set, defaults to Gson delegate).


asereda-gs:
@elucash sure take your time. 
I wanted to get your agreement (opinion) on general direction initially (there are still some cleanups to be done).

mongo/src/org/immutables/mongo/bson4gson/BsonReader.java
```diff
@@ -72,6 +73,26 @@ public void close() throws IOException {
     this.delegate = checkNotNull(delegate, "delegate");
   }
 
+  @Override
+  public <T> T unwrap(Class<T> type) {
```
elucash: thinking of if `Wrapper<T>` would be more staightforward and replace  `<T> T unwrap(Class<T> type)` with simple `T unwrap()`. For example,
`class BsonReader extends JsonReader implements Wrapper<org.bson.BsonReader> {`
then if you cast to `BsonReader` in adapter you know that `unwrap()` will get `org.bson.BsonReader`, if you don't cast, but `instanceof Wrapper<?>`, then `unwrap()` will return `Object` and the client will try to instanceof / cast on his own. Also, is it realistic that a delegate itself would be a wrapper in the way we need to care about? I just have a feeling that this piece is too flexible/promising, but actual usages are not much needing this.

asereda-gs: That's a fair point. 
Wrapper interface was inspired by [JDBC Wrapper](https://docs.oracle.com/javase/7/docs/api/java/sql/Wrapper.html). Perhaps it allows too much (eg. `unwrap(Closeable.class)`)

How about calling it `Delegate<T>` or re-using [ForwardingObject](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/ForwardingObject.html) ?

elucash: > Wrapper interface was inspired by JDBC Wrapper. Perhaps it allows too much (eg. unwrap(Closeable.class))

Yep, recognized patter. JDBC do not have a chance provide any type safety via `Wrapper<T>`, so they also decided that it's possible to get different kinds of delegates depending of a passed in class, so they decided to be quite flexible.

> How about calling it Delegate<T> or re-using ForwardingObject (in fact delegate() method is protected and doesn't help much) ?

`Wrapper<T>` or `DelegatesTo<T>` would be just fine, `ForwardingObject` is for their (Guava) needs.

> In general, do you agree with exposing original BsonReader (new way) instead of creating delegate methods (old way) ?

probably yes, don't see much overwhelming benefits on either side, probably less delegation in a new model

asereda-gs: Addressed your comments

mongo/src/org/immutables/mongo/repository/RepositorySetup.java
```diff
@@ -136,7 +182,37 @@ public Builder database(MongoDatabase database) {
...
+    }
+
+    /**
+     * Register custom property naming strategy. Allows to rename POJO properties
+     * during query construction.
+     *
+     * @param fieldNamingStrategy instance to be used during querying
+     * @return {@code this}
+     */
+    Builder fieldNamingStrategy(@Nullable FieldNamingStrategy fieldNamingStrategy) {
```
elucash: either `@Nullable` is accidental or I don't understand )

asereda-gs: It is accidental

mongo/src/org/immutables/mongo/repository/RepositorySetup.java
```diff
@@ -55,13 +63,46 @@
...
+                          FieldNamingStrategy fieldNamingStrategy) {
     this.executor = executor;
     this.database = database;
-    this.gson = gson;
+    this.codecRegistry = codecRegistry;
+    this.fieldNamingStrategy = Preconditions.checkNotNull(fieldNamingStrategy, "fieldNamingStrategy");
+  }
+
+  /**
+   * Allows to customize how attributes are named in mongo queries (this is different from document
```
elucash: I maybe still don't deep enough here to understand all implications, just sharing some observation: if we cannot align them automatically, is there any chance to explain it better in the javadoc. At first I thought there could be some kind of hack technique to align (field naming in query and in document) automatically. As a late though (I know we initially discussed it, but just in case), would it make any sense if translateName would belong to codec as a strong reminder that it should be implemented in coded to be auto-aligned between documents and queries? thoughts?

mongo/src/org/immutables/mongo/repository/RepositorySetup.java
```diff
@@ -136,7 +182,37 @@ public Builder database(MongoDatabase database) {
...
+
+    /**
+     * Register custom codec registry to serialize entities externally.
+     *
+     * @param registry existing registry
+     * @return {@code this}
+     */
+    Builder codecRegistry(CodecRegistry registry) {
+      this.codecRegistry = checkNotNull(registry, "registry");
+      this.fieldNamingStrategy = null;
```
elucash: troubling as missing up with order of `.fieldNamingStrategy` and `.codecRegistry` could leave it uninialiazed and even if documented, would possibly be one of those wtf moments. btw, this is one more hint to think about somehow merging field naming into codec.

mongo/src/org/immutables/mongo/repository/RepositorySetup.java
elucash: oh, I see codec registry is coming from driver. Even don't know... Maybe just document it and put field naming strategy as overload to add codec

```
.codecRegistry(registry, namingStrategy)
.codecRegistry(registry) // <-- use default naming strategy
```

mongo/src/org/immutables/mongo/repository/RepositorySetup.java
asereda-gs: Yes. I'm not a big fan of separate `FieldNamingStrategy` myself. Perhaps expose something like:
```java
class Filter  {
    final Member member;
    final Object value;
    final QueryOperators op; // optional  
}
```
This information can be saved during query construction.

elucash: Than would be definitely moving toward more comprehensive solution, but I think we're still far enough from really optimal solution. Given that we reuse codec registries from mongo driver, I feel this should be acceptable compromise to have separate naming strategy as we currently have in this PR.

I think we just need

* More elaborately explain in Javadoc why naming strategy needed and why it works only for queries, and how codec itself should use the same strategy
* More elaborately explain or at least hint in Javadoc of `translateName` how it supposed to be implemented (the example could be Jackson and that translate name is supposed to read `@JsonProperty` annotation or reach out to Jackson's naming strategies). An example (or hint) is, probably, not required, but nice to have
* Set codec registry and naming strategy at the same time, and in javadoc give yet another reminder and pointer to read strategy's javadoc

```
.codecRegistry(codecRegistry, namingStrategy)
```

asereda-gs: What would be an optimal solution in your view ? I would like to avoid introducing new interface if there is a better alternative.

So far I see the following options:
1. Explicit `FieldNamingStrategy` interface. I don't like it because it introduces new API.
2. Bson `Encoder<Field/Member/Method>` or `Encoder<Filter>` (where `Filter` is specific class to immutables)

Obviously javadoc has to be improved.

elucash: I'm kinda ok with `1.` although fully share your dislike it introducing APIs
About `2.` I guess I get where it's heading, but not fully understand it yet, so cannot be sure

asereda-gs: I've added more explanations to javadoc. Let me know what you think.

Additional question:
In `Repository.generator` [you're generating](https://github.com/immutables/immutables/blob/master/value-processor/src/org/immutables/value/processor/Repositories.generator#L828) separate class with same fields. Any reason why not get members from existing interface or generated immutables class  via same reflection ?

elucash: as far as I can remember, we cannot always rely on a field to be present, some attributes can be implemented in a special way. In my view, this is just a peculiarity of Gson API to require `Field` object as input. So the current way seems easy and reliable way to get those fields. I've tried to construct some field objects in support library but that was fragile (some com sun unsafe stuff etc). And I don't know if there's a way out of this

asereda-gs:
@elucash please let me know if this version looks OK to you

elucash:
@asereda-gs  looks really good, excellent job! One place I still trying to wrap my head around is when we create `newGson`, but I hope I will get it. Just will make one more glimpse, overall I feel it's good to go

asereda-gs:
@elucash Yes that was the tricky part. I didn't want existing Gson users to be forced to modify their code with this change. In the same time, Gson is becoming "second class" citizen since serialization is now delegated to [Codecs](http://mongodb.github.io/mongo-java-driver/3.8/bson/codecs/) abstraction for both queries (aka Criteria/Constraint) and POJOs. Having both worlds wasn't straight-forward. 
 
The main problem is defining correct priority for Gson and Bson adapters (when they co-exists) depending on the type. For instance `Date` should be serialized by Gson (even if Bson has codec for it). But `ObjectId` / `Document` or `BsonDocument` by Bson.

My biggest worry is that this feature doesn't change the "wire format". Ie existing persisted data can be read and written as before. Mongo database shouldn't "notice" any difference.

asereda-gs:
@elucash I might have more refinements as part of  #835  (jackson for mongo adapter)

PR #843, opened 2018-10-01
Feature/optionals factory

### LINKED ISSUE #800: Allow to pass type T in constructor when using parameter Optional<T>
Hello,

First of all, thanks for the library, it is really useful.

I'm currently working with the version 2.6.1, and I customized Immutables annotation to have a classic constructor

``` java
@Value.Style(
        get = {"is*", "get*"},
        init = "set*",
        of = "new",
        typeAbstract = {"*Definition"},
        typeImmutable = "*",
        visibility = Value.Style.ImplementationVisibility.PUBLIC,
        allParameters = true,
        jdkOnly = true,
        defaults = @Value.Immutable(copy = false, builder = false))
```

However, when I use Optional as a return type, I'm obliged to provide and Optional in the constructor because the class

``` java
interface Foo {
  Optional<Bar> getBar();
}
```

Will lead to the generation of this constructor

``` java
public Foo(Optional<? extends Bar> bar) {
  this.bar = bar.orElse(null);
}
```

What I would like to have is this constructor

``` java
public Foo(@Nullable Bar bar) {
  this.bar = bar;
}
```

Is there a way to generate such a constructor ?

ThLeu:
Hey!

This PR closes #800 

I tried my luck at the topic in Issue #800 and added the generation of additional static factory methods and constructors for Optional types. This allows the use of the unwrapped type in addition to the current implementation which requires the variable to be wrapped in an Optional when constructing the Immutables object.

Please have a look and tell me what you think and if there are any issues with the implementation or anything.

Best regards

### DIFF
--- a/value-annotations/src/org/immutables/value/Value.java
+++ b/value-annotations/src/org/immutables/value/Value.java
@@ -809,7 +809,7 @@ public @interface Value {
     ImplementationVisibility visibility() default ImplementationVisibility.SAME;
 
     /**
-     * Specify whether init and copy methods for an unwrapped {@code X} of {@code Optional<X>}
+     * Specify whether init, copy and factory methods and constructors for an unwrapped {@code X} of {@code Optional<X>}
      * should accept {@code null} values as empty value. By default nulls are rejected in favor of
      * explicit conversion using {@code Optional.ofNullable}. Please note that initializers that
      * take explicit {@code Optional} value always reject nulls regardless of this setting.
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -2231,11 +2231,13 @@ return [type.factoryOf]([output.linesShortable][for v in type.settableAttributes
 [/if]
 [/output.trim][/template]
 
-[template valueFromValueNoOptional Attribute v String expression][output.trim]
+[template valueFromValueNoOptional Type type Attribute v String expression][output.trim]
 [if v.encoding]
   [valueFromValue v expression]
+[else if v.jdkOptional and (v.jdkSpecializedOptional or v.optionalAcceptNullable)]
+  [maybeCopyOf v][expression][/maybeCopyOf]
 [else if v.jdkOptional]
-  [expression]
+  [maybeCopyOf v][requireNonNull type]([expression], "[expression]")[/maybeCopyOf]
 [else if v.optionalType]
   [optionalOf v]([expression])
 [else]
@@ -2446,18 +2448,18 @@ this.[n] = [valueFromValue v][invokeSuper v].[v.names.get]()[/valueFromValue];
   [let shim][disambiguateField type 'initShim'][/let]
   [for v in type.constructorArguments, n = v.name]
 [if v.generateDefault and type.generateSafeDerived]
-    [shim].[v.names.init]([valueFromValueNoOptional v n]);
+    [shim].[v.names.init]([valueFromValueNoOptional type v n]);
 [else if v.hasVirtualImpl]
-    [rr.virtualImpl v] = [valueFromValueNoOptional v n];
+    [rr.virtualImpl v] = [valueFromValueNoOptional type v n];
 [else]
-    this.[n] = [valueFromValueNoOptional v n];
+    this.[n] = [valueFromValueNoOptional type v n];
 [/if]
   [/for]
     [generateConstructorDefaultAttributes type type.constructorOmited]
     [generateDerivedConstruction type type.constructorArguments]
 [else][-- Alternatively a lot simpler implementation]
   [for v in type.constructorArguments, n = v.name]
-    this.[n] = [valueFromValueNoOptional v n];
+    this.[n] = [valueFromValueNoOptional type v n];
   [/for]
 [/if]
     [generateAfterConstruction type false]
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -702,9 +702,9 @@ public [type.typeAbstract.relative] [v.names.with]([v.atNullability][v.type] val
 
   /**
    * Construct a new immutable {@code [type.name]} instance.
-  [for v in type.constructorArguments]
+[for v in type.constructorArguments]
    * @param [v.name] The value for the {@code [v.name]} attribute
-  [/for]
+[/for]
    * @return An immutable [type.name] instance
    */
   public static [type.generics.spaceAfter][type.typeValue] [type.names.of]([for v in type.constructorArguments][if not for.first], [/if][v.atNullability][eachLine v.constructorParameterInjectedAnnotations][v.type] [v.name][/for]) {
@@ -714,21 +714,21 @@ public [type.typeAbstract.relative] [v.names.with]([v.atNullability][v.type] val
 
   /**
    * Construct a new immutable {@code [type.name]} instance.
-  [for v in type.constructorArguments]
+[for v in type.constructorArguments]
    * @param [v.name] The value for the {@code [v.name]} attribute
-  [/for]
+[/for]
    * @return An immutable [type.name] instance
    */
   public static [type.generics.spaceAfter][type.typeValue] [type.names.of]([for v in type.constructorArguments][if not for.first], [/if][v.atNullability][eachLine v.constructorParameterInjectedAnnotations][constructorAcceptType v] [v.name][/for]) {
     return [validated type false]new [type.typeImmutable.relativeRaw][type.generics.diamond]([for v in type.constructorArguments][if not for.first], [/if][v.name][/for])[/validated];
   }
-
   [if type.hasOptionalConstructorArguments]
+
   /**
    * Construct a new immutable {@code [type.name]} instance.
-  [for v in type.constructorArguments]
+[for v in type.constructorArguments]
    * @param [v.name] The value for the {@code [v.name]} attribute
-  [/for]
+[/for]
    * @return An immutable [type.name] instance
    */
   public static [type.generics.spaceAfter][type.typeValue] [type.names.of]([for v in type.constructorArguments][if not for.first], [/if][v.atNullability][eachLine v.constructorParameterInjectedAnnotations][if v.isOptionalType][unwrappedOptionalType v][else][constructorAcceptType v][/if] [v.name][/for]) {
@@ -2381,8 +2381,8 @@ this.[n] = [valueFromValue v][invokeSuper v].[v.names.get]()[/valueFromValue];
     [generateAfterConstruction type false]
   [/output.collapsible]}
 [/if]
-
 [if type.useConstructor andnot type.generateConstructorUseCopyConstructor]
+
   [if type.factoryOf.new]
   /**
    * Construct a new immutable {@code [type.name]} instance.
@@ -2420,9 +2420,9 @@ this.[n] = [valueFromValue v][invokeSuper v].[v.names.get]()[/valueFromValue];
 [/if]
     [generateAfterConstruction type false]
   }
-
 [-- Additional constructors for Optional types --]
 [if type.hasOptionalConstructorArguments]
+
   [if type.factoryOf.new]
   /**
    * Construct a new immutable {@code [type.name]} instance.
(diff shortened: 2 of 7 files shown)


elucash:
Thank you for the pull request! There are (will add) some review comments inline

value-annotations/src/org/immutables/value/Value.java
```diff
@@ -809,7 +809,7 @@
     ImplementationVisibility visibility() default ImplementationVisibility.SAME;
 
     /**
-     * Specify whether init and copy methods for an unwrapped {@code X} of {@code Optional<X>}
+     * Specify whether init, copy and factory methods and constructors for an unwrapped {@code X} of {@code Optional<X>}
```
elucash: I see we're supposed to cover this features under existing feature flag `optionalAcceptNullable`, which is good in my opinion. However, (plz correct if I'm wrong) when we generating code we only ask about `type.hasOptionalConstructorArguments` but we're not querying this `optionalAcceptNullable` style attribute, thus we always generate additional factory methods if we have at least one optional attribute, which is not good

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -702,26 +702,52 @@ public [type.typeAbstract.relative] [v.names.with]([v.atNullability][v.type] val
 
   /**
    * Construct a new immutable {@code [type.name]} instance.
-[for v in type.constructorArguments]
+  [for v in type.constructorArguments]
```
elucash: I'm not against indenting `for` statement with additional 2 spaces here, but those was "outdented" on purpose so they are stick out a bit. I use this ofter in the template: combination of increasing and decreasing indentation where it might help readability. "Fixing" it here create inconsistency with how it is made in all other [untouched] places, so improvement is questionable.

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -702,26 +702,52 @@ public [type.typeAbstract.relative] [v.names.with]([v.atNullability][v.type] val
...
    * @param [v.name] The value for the {@code [v.name]} attribute
-[/for]
+  [/for]
    * @return An immutable [type.name] instance
    */
   public static [type.generics.spaceAfter][type.typeValue] [type.names.of]([for v in type.constructorArguments][if not for.first], [/if][v.atNullability][eachLine v.constructorParameterInjectedAnnotations][v.type] [v.name][/for]) {
     return [type.names.of]([for v in type.constructorArguments][if not for.first], [/if][if v.requiresAlternativeStrictConstructor]([constructorAcceptType v]) [/if][v.name][/for]);
   }
+
```
elucash: Here and in other places, we need to refine so the newline appears inside the guarding `if` directive. The reason for this is that when if statement is not executing (having false condition), we will not leave extra unnecessary newlines in generated code. As generated code prettiness is valuable for us, we follow the convention: always put newline _before_ generated methods or nested class, but _inside_ guarding `if` or `for` or `template` directives, like this

```java
[if condition]

/**
 * Method
 */
void method() {
  return;
}
[/if]
[if condition andnot condition2]

interface NestedAndUseless {}
[/if]
```

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -702,26 +702,52 @@ public [type.typeAbstract.relative] [v.names.with]([v.atNullability][v.type] val
...
+  [for v in type.constructorArguments]
    * @param [v.name] The value for the {@code [v.name]} attribute
-[/for]
+  [/for]
    * @return An immutable [type.name] instance
    */
   public static [type.generics.spaceAfter][type.typeValue] [type.names.of]([for v in type.constructorArguments][if not for.first], [/if][v.atNullability][eachLine v.constructorParameterInjectedAnnotations][constructorAcceptType v] [v.name][/for]) {
     return [validated type false]new [type.typeImmutable.relativeRaw][type.generics.diamond]([for v in type.constructorArguments][if not for.first], [/if][v.name][/for])[/validated];
   }
+
```
elucash: Here and in other places, we need to refine so the newline appears inside the guarding `if` directive. The reason for this is that when if statement is not executing (having false condition), we will not leave extra unnecessary newlines in generated code. As generated code prettiness is valuable for us, we follow the convention: always put newline _before_ generated methods or nested class, but _inside_ guarding `if` or `for` or `template` directives, like this

```java
[if condition]

/**
 * Method
 */
void method() {
  return;
}
[/if]
[if condition andnot condition2]

interface NestedAndUseless {}
[/if]
```

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -702,26 +702,52 @@ public [type.typeAbstract.relative] [v.names.with]([v.atNullability][v.type] val
...
    * @param [v.name] The value for the {@code [v.name]} attribute
-[/for]
+  [/for]
    * @return An immutable [type.name] instance
    */
   public static [type.generics.spaceAfter][type.typeValue] [type.names.of]([for v in type.constructorArguments][if not for.first], [/if][v.atNullability][eachLine v.constructorParameterInjectedAnnotations][v.type] [v.name][/for]) {
     return [type.names.of]([for v in type.constructorArguments][if not for.first], [/if][if v.requiresAlternativeStrictConstructor]([constructorAcceptType v]) [/if][v.name][/for]);
   }
+
+  [if type.hasOptionalConstructorArguments]
```
elucash: suggest to remove this variation, I think we don't need to mess with special `type.requiresAlternativeStrictConstructor` mode here (we're inside this condition). Unless this is really needed for something, any examples? The initial idea was that we use in general `Iterable<? extends Something>` in params for collection, but for compatibility (was needed at some point, I guess) we also adding constructor where those are `List<Something>` originally declared in attribute. You can try mixing with constructor arguments of new optional with those generic lists and see if we can get duplicate or otherwise unnecessary constructors. In general I consider these new (which accept nullable for optional) constructors as anything but strict so would prefer to not multiply such combinations.

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -2395,9 +2433,49 @@ this.[n] = [valueFromValue v][invokeSuper v].[v.names.get]()[/valueFromValue];
...
+   */
+  [eachLine type.constructorAnnotations]
+  [eachLine type.constructorInjectedAnnotations]
+  [suppressAnyCovariantCastForOptional type.constructorArguments]
+  public [type.typeImmutable.simple]([output.linesShortable][for v in type.constructorArguments][if not for.first],[/if]
+      [v.constructorParameterAnnotations][v.atNullability][if v.isOptionalType][unwrappedOptionalType v][else][constructorAcceptType v][/if] [v.name][/for][/output.linesShortable]) {
+  [else]
+  [suppressAnyCovariantCastForOptional type.constructorArguments]
+  private [type.typeImmutable.simple]([output.linesShortable][for v in type.constructorArguments][if not for.first],[/if]
+      [v.atNullability][if v.isOptionalType][unwrappedOptionalType v][else][constructorAcceptType v][/if] [v.name][/for][/output.linesShortable]) {
```
elucash: need to add `@Nullable` (or equivalent) where it is substitute for `Optional`. Here I assume we only propagate nullable if it's declared nullable, but I might be wrong here. Have you checked those nullable annotations?

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -2395,9 +2433,49 @@ this.[n] = [valueFromValue v][invokeSuper v].[v.names.get]()[/valueFromValue];
...
+  [for v in type.constructorArguments, n = v.name]
+    this.[n] = [valueFromValueNoOptional v n];
+  [/for]
+[/if]
+    [generateAfterConstruction type false]
+  }
+[/if]
 [/if]
 [if type.generateBuilderConstructor]
-
```
elucash: this blank line is needed, see the previous comment

value-annotations/src/org/immutables/value/Value.java
ThLeu: I think what I wanted to do was to add a `Objects.requireNonNull()` to the corresponding constructor when the flag was not true. But it seems I totally forgot about that in the end, sorry about that. 
In my opinion this would make more sense as it would behave the same as the generated with (copy) methods.
What do you think, would it be ok if I added the `Objects.requireNonNull()`?

value-processor/src/org/immutables/value/processor/Immutables.generator
ThLeu: Sorry, I shouldn't have pushed that as it is probably just a remainder of when I tried to figure everything out, I will remove it again.

value-processor/src/org/immutables/value/processor/Immutables.generator
ThLeu: Hm...ok I didn't realise that this was the idea behind it, I'll change it accordingly.

value-processor/src/org/immutables/value/processor/Immutables.generator
ThLeu: Right, I will remove it, I figured that it should be there for the same reason that the current alternative is there but I have no example. If it would make sense it can be added later.

value-processor/src/org/immutables/value/processor/Immutables.generator
ThLeu: As far as i saw this behaves as expected according to the parameter `optionalAcceptNullable` if it is true `@Nullable` is added otherwise it isn't.
If I also correct and add `Objects.requireNonNull()` (see first comment) it would behave correctly in my opinion.
The `atNullable` is called inside of `unwrappedOptionalType`.

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -2395,9 +2433,49 @@ this.[n] = [valueFromValue v][invokeSuper v].[v.names.get]()[/valueFromValue];
 [/if]
     [generateAfterConstruction type false]
   }
+
+[-- Additional constructors for Optional types --]
+[if type.hasOptionalConstructorArguments]
```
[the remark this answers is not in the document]
ThLeu: Yes, this was definitely the trickiest part and I am also not really happy with copy-pasting it, but in the end it seemed to be the safest / most readable option for now as I do not know the ins and outs of the template language and it's possibilities.
But if you give me a hint on how to pass some invokable fragments like here the `valueFromValue` or `valueFromValueNoOptional`, I will gladly change it.

ThLeu:
Thank you for the thorough review!
I added some comments and made additional changes according to your comments.

elucash:
Thank you for the changes! Now it looks good. We only need to either document the copy-paste of the factory code as I've mentioned. Or if you want to minimize copy-paste and dive a bit in hacking with the arcane template engine, here's some hints:

* Templates can be captured as instances of invokable and passed to other templates, so templates can apply arguments to it
* To declare template parameter accepting invokable fragment, we need to declare type of it. `Immutables.generator` doesn't have this defined, but, for example `Renderers.generator` has.

```java
// inside class Immutables (Immutables.java code-behind class)
  @Generator.Typedef
  Templates.Invokable Invokable; // field name will be the type name used in template
```

* Then in template create/use regular and higher order templates

```
[template x String param]
X: [param]
[/template]
[template y String param]
Y: [param]
[/template]

[template higherOrder Invokable a Invokable b]
[a 'AAA'] [-- use invokable just by applying params (if any)--]
  [b 'BBB'] [-- params types and count are not compile-checked so AIOO or CCE possible
[/template]

[template usingTemplate]
[higherOrder x y]
[higherOrder y x]
[/template]
```

the expected output of `usingTemplate` would be 

```
X: AAA
  Y: BBB
Y: AAA
  X: BBB
```

* can use `[output.trim] ... [/output.trim]` around template contents if surrounding newlines and whitespaces are to be collapsed

* the other way is just create template accepting condition `Boolean useOptionalAcceptNullable`
and have some more `[if]/`[else]` directives to output working in one or another mode, `[let x]content[/let]` can be used to capture some fragments as "local variables", where conditions can also be places inside such `let`

Thank you!

ThLeu:
Thank you for your helpful comment about invokable fragments.
I did now remove the redundant code (see last commit), what do you think?

elucash:
@ThLeu I looks really nice and should work fine, I'll take last round of review to check some details and will merge tomorrow. Thank you for the hard work!

ThLeu:
Thank you, if you run into any issues I'll have a look at it.

elucash:
@ThLeu I'm merging other stuff now, so I can then resolve the conflict and merge this PR. But we're ready to go here, thanks!

PR #844, opened 2018-10-02
Support protocol buffer attributes

daicoden:
The discovery mechanism needs to be more strict for protocol buffers
    because there are several methods which have the right signature, but
    are not part of the builder API.

    First, we make sure all methods for an attribute builder match the
    specified Value#attributeBuilder patterns

    Second, we ensure that the return methods are equivalent using
    typeMirriors#isSameType. Using equals was working intermittently, and
    the documentation on equals says to use the isSameType instead.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/meta/AttributeBuilderReflection.java
+++ b/value-processor/src/org/immutables/value/processor/meta/AttributeBuilderReflection.java
@@ -14,6 +14,8 @@ import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
+import javax.lang.model.util.Types;
+
 import org.immutables.value.Value.Derived;
 import org.immutables.value.Value.Immutable;
 import org.immutables.value.Value.Lazy;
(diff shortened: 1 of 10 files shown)


value-processor/src/org/immutables/value/processor/meta/AttributeBuilderReflection.java
```diff
@@ -339,12 +341,19 @@ private static boolean isPossibleBuildMethod(ValueAttribute valueAttribute, Elem
...
+      Types typeUtils
+          = valueAttribute.containingType.constitution.protoclass().environment().processing().getTypeUtils();
+
       ExecutableElement candidateBuildMethod = (ExecutableElement) possibleBuildMethod;
       return !candidateBuildMethod.getModifiers().contains(Modifier.STATIC)
           && candidateBuildMethod.getModifiers().contains(Modifier.PUBLIC)
           && candidateBuildMethod.getTypeParameters().isEmpty()
           && candidateBuildMethod.getReturnType().getKind() == TypeKind.DECLARED
-          && (candidateBuildMethod.getReturnType()).equals(valueAttribute.containedTypeElement.asType());
+          && (typeUtils.isSameType(candidateBuildMethod.getReturnType(), valueAttribute.containedTypeElement.asType()));
```
elucash: I wonder what we've use elsewhere (see `FromSupertypesModel.isEligibleFromType` which might be overkill..) In any case we need to use `CachingElements.getDelegate` before passing these types/elements to any of `Types`/`Elements` util methods as we can have wrapped instances and Javac may throw exception (even if this is not happening for some reason, but for consistency)

elucash:
Thank you for the PR, I'll need couple of days to get thru it, I'll leave some more feedback

elucash:
looks ok, merged. Thank you!

PR #845, opened 2018-10-05
Improve `from` behavior for mandatory fields on partial Modifiable instances

stevenschlansker:
Currently the Modifiable implementations check if a field has been set and throw if not, which prevents a partial Modifiable object seeding a new builder.

This PR adds a Builder and Modifiable `from` overload for the Modifiable type, as well as a dynamic `instanceof` check since we can't rely on people statically calling the overload.  The overload then checks `isSet` on mandatory attributes before attempting to copy them, avoiding the exception.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -1253,6 +1253,7 @@ public interface BuildFinal[type.generics] {
   @SuppressWarnings("unchecked")
 [/if]
   private void from(Object object) {
+    [dynamicFromModifiableCheck type 'object' '']
     [for l in bs.positions.longs]
     [atVar type]long bits[emptyIfZero l.index] = 0;
     [/for]
@@ -1298,6 +1299,7 @@ public interface BuildFinal[type.generics] {
   [atCanIgnoreReturnValue type]
   [type.typeAbstract.access]final [builderReturnType type] [type.names.from]([type.typeAbstract.relative] instance) {
     [requireNonNull type](instance, "instance");
+    [dynamicFromModifiableCheck type 'instance' 'this']
     [for v in setters]
     [buildFromAttribute v]
     [/for]
@@ -1307,8 +1309,25 @@ public interface BuildFinal[type.generics] {
   [atCanIgnoreReturnValue type]
   public final [builderReturnType type] [type.names.from]([type.typeAbstract.relative] instance) {
     [requireNonNull type](instance, "instance");
+    [dynamicFromModifiableCheck type 'instance' 'this']
     [for v in setters]
     [buildFromAttribute v]
+    [/for]
+    return [builderReturnThis type];
+  }
+  [/if]
+  [if type.kind.isModifiable]
+
+  /**
+   * Fill a builder with attribute values from the provided {@code [type.typeModifiable]} instance.
+   * @param instance The instance from which to copy values
+   * @return {@code this} builder for use in a chained invocation
+   */
+  [atCanIgnoreReturnValue type]
+  public final [builderReturnType type] [type.names.from]([type.typeModifiable] instance) {
+    [requireNonNull type](instance, "instance");
+    [for v in setters]
+    [buildFromMandatoryAttribute buildFromAttribute v 'instance']
     [/for]
     return [builderReturnThis type];
   }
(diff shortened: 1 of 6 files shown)


elucash:
Thank you @stevenschlansker for your work! I think that having that kind of code should be quite a good contribution. Let us know when you will have something you're comfortable to review.
As a hint, if you need to share some templates between `Immutables.generator` and `Modifiables.generator`, you can declare them in `Immutables.generator` as `[template public name.....]` then add corresponding signature in `Immutables.java` as `abstract Templates.Invokable name....();` and this will enable to use it as `[im.name..` in `Modifiables.generator` as it's already "imported" there using `im.` prefix

stevenschlansker:
Thanks @elucash for the hint, I was able to factor out the dynamic check.  I wanted to factor out the `if(isset) { buildFromAttribute }` block, but unfortunately `buildFromAttribute` is defined differently in Immutables vs Modifiables, so factoring that out of Modifiables causes it to use the wrong method names from Immutables.  Is it allowed to pass a template (function) as data?  i.e.
```
[template public doSomething Invokable attributeBuilder Attribute v]
[attributeBuilder v]
[/template]

[doSomething buildFromAttribute v]
```
If not, doing further factoring of this PR may require reorganizing existing code, which could be a followup improvement :)

elucash:
@stevenschlansker it is possible to pass fragments/templates to other templates using Invokable as a parameter, here's example https://github.com/immutables/immutables/pull/843#issuecomment-428390972

stevenschlansker:
This template engine is arcane indeed!  I will give that an attempt.

stevenschlansker:
That wasn't too bad!  @elucash I think this version is ready for review.

value-processor/src/org/immutables/value/processor/Modifiables.generator
```diff
@@ -401,11 +401,31 @@ private void from(Object object) {
...
+ * Any of the instance's absent optional values will not be copied (will not override current values).
+[if type.hasSettableCollection or type.hasSettableMapping]
+ * Collection elements and entries will be added, not replaced.
+[/if]
+ * @param instance The instance from which to copy values
+ * @return {@code this} for use in a chained invocation
+ */
+public [thisReturnType type] [type.names.from]([type.typeModifiable] instance) {
+  [im.requireNonNull type](instance, "instance");
+  final [type.typeModifiable] cast = ([type.typeModifiable]) instance;
```
elucash: don't need to cast here, param already modifiable

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -1307,11 +1309,27 @@ public interface BuildFinal[type.generics] {
...
     [requireNonNull type](instance, "instance");
+    [dynamicFromModifiableCheck type 'instance' 'this']
     [for v in setters]
     [buildFromAttribute v]
     [/for]
     return [builderReturnThis type];
   }
+  [/if]
+  [if type.kind.isModifiable]
+  /**
```
elucash: probably need to add a blank line inside `[if` before the method. For methods, nested types and sometimes group of fields, when we have conditional output we put separating blank lines inside `[if` or `[for` directives, but before the methods so generated code would have needed blank lines and will not contain extra blank lines if optional code generated and when not.

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -1253,6 +1253,7 @@ public interface BuildFinal[type.generics] {
   @SuppressWarnings("unchecked")
 [/if]
   private void from(Object object) {
+    [dynamicFromModifiableCheck type 'object' '']
```
elucash: `return ;` with extra space would be a little ugly here, maybe add some conditions inside `dynamicFromModifiableCheck`, like `return[if returnValue] [returnValue][/if];`

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -2479,6 +2497,25 @@ this.[n] = [valueFromValue v][invokeSuper v].[v.names.get]()[/valueFromValue];
 [/for]
 [/template]
 
+[template public buildFromMandatoryAttribute Invokable buildAttribute Attribute v String varName]
+[if v.isMandatory]
+if ([varName].[v.names.isSet]()) {
+  [buildAttribute v]
+}
+[else]
+  [buildAttribute v]
```
elucash: is this extra indentation, please, check generated source so it have proper formatting for both with `if ({`  and without

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -2479,6 +2497,25 @@ this.[n] = [valueFromValue v][invokeSuper v].[v.names.get]()[/valueFromValue];
...
+[/if]
+[/template]
+
+[template public dynamicFromModifiableCheck Type type String varName String returnValue]
+[if type.kind.isModifiable]
+if ([varName] instanceof [type.typeModifiable.relativeRaw]) {
```
elucash: maybe as a matter of style/our preference we should add `[type.generics.unknown]` after `.relativeRaw]` so if type have generic parameters, it will output trailing `<?>` after raw type name (we try to avoid any use or raw types, even in instanceof I prefer to check it agains generic type with unknown arguments).
This also prompts to have at least one compilation test for value type having generic type parameters. We need check that we don't have new unsuppressed unchecked cast warnings in generated code.

elucash:
Thank you @stevenschlansker, really good work! there are some minor review comments to check, but overall we're good to go

stevenschlansker:
@elucash thanks for the review, I believe I've addressed all of your comments.  I also checked the indentation on the generated builder and it seems OK.

elucash:
Looks really cool! Thank you @stevenschlansker ! merging

Issue #603, opened 2017-04-21
Immutable with Jackson annotation @JsonIgnoreProperties with ignore properties failing during deserialization not working  


shyam4u:
If I have an immutable which has a derived field having no setter and I want to ignore the same field during de- serialization am unable to do so .I keep getting unrecognized property exception from jackson .
org.codehaus.jackson.map.exc.UnrecognizedPropertyException:

Below is the sample immutable and my used case 

@JsonIgnoreProperties(value={ "computed" }, allowGetters=true)
@Value.Immutable
@Value.Style(typeImmutable = "MyClassImpl", jdkOnly = true)
@JsonTypeName("myClass")
@JsonDeserialize(builder = MyClassImpl.Builder.class)
public abstract class  MyClass {

	@Value.Derived
	@JsonProperty("type")
	public String getType() {
		return "CATEGORY";
	}
	
	@JsonProperty("name")
	public String getName()
}

I guess we should have the derived property ignored by default while deserialization .We need the serialization to happen as it is a read only property.Is there any way we can achieve this feature using immutable ?

elucash:
Hi, thank you for reporting this. I would gladly try to reproduce the issue, but it seems that you have very old version of Jackson running (1.x) judging from the `org.codehaus.jackson.map.exc.UnrecognizedPropertyException` fully qualified name. Immutables only works with Jackson version 2.5 or higher. Please, untangle the dependencies and let us know if the problem persists.

shyam4u:
No @elucash  I have not pasted the correct stack trace .My version of jackon is 2.5+ and immutables is 2.4+

augustotravillio:
@shyam4u Have you tried `@JsonIgnoreProperties(ignoreUnknown = true)`

ldriscoll:
@augustotravillio @elucash I've also run into something like this.
Basically, `@Derived` causes us to not have any setters for that value in the `Json` class.  The case I've seen is where the getter is for a List; so I think Jackson is calling the getter, to get the list, to add new values to it; and we get the `UnsupportedOperationException`.

`@JsonIgnoreProperties` is not helping in this case.

I've tried annotating the method as
```java
@Value.Derived
@JsonProperty(access = Access.READ_ONLY)
public List<String> getThings() {
...
```
and this also doesn't work. 

Basically, we need a way to have `@Value.Derived` that can be serialised, but not fail during deserialisation (since we have javascript front ends that need the data there)

jaymmodi:
[...]

PR #846, opened 2018-10-09
Issue #603 - automatically add `@JsonIgnore` to the `class Json` builder for derived functions.

jdstapleton:
Since derived functions are calculated fields that are outputted when serialized to json, they
shouldn't be attempted to be deserialized from the Json structure.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -2670,6 +2670,9 @@ static final [output.linesShortable]class Json[type.generics]
   [signature] { throw new UnsupportedOperationException(); }
 [/for]
 [for v in type.allAccessibleAttributes]
+  [if v.isGenerateDerived]
+  @com.fasterxml.jackson.annotation.JsonIgnore
+  [/if]
   @Override[-- we actually don't need this implementation, signature is good enough --]
   [v.toSignature] { throw new UnsupportedOperationException(); }
 [/for]
--- a/value-processor/src/org/immutables/value/processor/meta/ValueAttribute.java
+++ b/value-processor/src/org/immutables/value/processor/meta/ValueAttribute.java
@@ -581,6 +581,10 @@ public final class ValueAttribute extends TypeIntrospectionBase implements HasSt
         && !containingType.serial.isSimple();
   }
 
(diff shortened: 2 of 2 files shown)


elucash:
thank you for the PR! please see inline comments

value-processor/src/org/immutables/value/processor/meta/ValueAttribute.java
```diff
@@ -581,6 +581,10 @@ public boolean isGenerateTransientDerived() {
         && !containingType.serial.isSimple();
   }
 
+  public boolean isGenerateDerived() {
```
elucash: the getter is not needed here, field works just fine

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -2670,6 +2670,9 @@ static final [output.linesShortable]class Json[type.generics]
   [signature] { throw new UnsupportedOperationException(); }
 [/for]
 [for v in type.allAccessibleAttributes]
+  [if v.isGenerateDerived]
```
elucash: should we mark all stub methods as ignored (not only derived), does it cause any problem with deserialization of mandatory attributes?

jdstapleton: I don't know for sure. 

I think it only matters, if they are whatever Jackson considers 'getters' in this case.  Since the `class Json` is for deserialization.

https://fasterxml.github.io/jackson-annotations/javadoc/2.5/com/fasterxml/jackson/annotation/JsonIgnore.html

I don't know enough about the lazy attributes, I haven't used them, if they need it or not.

value-processor/src/org/immutables/value/processor/meta/ValueAttribute.java
jdstapleton: I must have had something else wrong when I first modified the `Immutables.generator` file.  But after rebuilding, it appears I didn't need to do that.

elucash: sorry, that crazy mix of fields and accessors grew "organically", just don't want to change some aspects of it (as trying to remember which ones computed up-front, which lazily and where). thank you for understanding

PR #851, opened 2018-10-18
Update copyright year to 2018 (in java and xml headers)

asereda-gs:
Change java and xml (pom.xml) headers to include current year


asereda-gs:
Perhaps it makes sense to add a checkstyle rule (in maven) for license header in all java / XML files ?

elucash:
I dunno, I initially updated only touched files, tried to delaying complicating this (pardon, automating) as long as I can. I'm ok with certain inconsistency there, it's just when new files are created with previous year(s) then it's something I can notice. As for checkstyle, the same, tried to not introduce it for as long as possible, and if introduce, it would be pretty strict )

PR #853, opened 2018-10-18
Update jackson-core version in gson module (2.8.5 -> 2.8.11)

asereda-gs:
Same version should be used across all modules


elucash:
Thank you!

Issue #827, opened 2018-09-12
The InitShim class is not annotated by `org.immutables.value.Generated`


matt-isaman:
It looks like the InitShim class was missed in the resolution to #756 and #792. I am using Immutables 2.7.1 and have an `@Immutable` object with a couple of `@Default` annotated getters. A special note here is that the InitShim class seems to only be generated if you have 2+ `@Default` methods.

```java
import org.immutables.value.Value.Default;
import org.immutables.value.Value.Immutable;

@Immutable
public abstract class Foo {

    public abstract String getString();

    @Default
    public Integer getI() {
        return 1;
    }

    @Default
    public Long getL() {
        return 1L;
    }
}
```
Generates this class (lots of code snipped out to only show the relevant pieces):
```java
/**
 * Immutable implementation of {@link Foo}.
 * <p>
 * Use the builder to create immutable instances:
 * {@code ImmutableFoo.builder()}.
 */
@SuppressWarnings({"all"})
@ParametersAreNonnullByDefault
@Generated("org.immutables.processor.ProxyProcessor")
@org.immutables.value.Generated(from = "Foo", generator = "Immutables")
@Immutable
@CheckReturnValue
public final class ImmutableFoo extends Foo {
  // <SNIP/>

  private final class InitShim {
    // <SNIP/>
  }

  // <SNIP/>

  /**
   * Builds instances of type {@link ImmutableFoo ImmutableFoo}.
   * Initialize attributes and then invoke the {@link #build()} method to create an
   * immutable instance.
   * <p><em>{@code Builder} is not thread-safe and generally should not be stored in a field or collection,
   * but instead used immediately to create instances.</em>
   */
  @org.immutables.value.Generated(from = "Foo", generator = "Immutables")
  @NotThreadSafe
  public static final class Builder {
    // <SNIP/>
  }
}
```
The InitShim, and any other internal class that Immutables may generate, should have the `@org.immutables.value.Generated(from = "Foo", generator = "Immutables")` annotation added.

before:
I've found more nested classes that need to be annotated with `org.immutables.value.Generated`.
* `*.InternProxy`
* `*.GsonAdapters*.*TypeAdapter`
* `*.*Repository.Finder`

elucash:
@before  Yep, I was a bit lazy to go over other generators, thank you for the PR!

PR #854, opened 2018-10-20
Add org.immutables.value.Generated annotation to nested classes #827

before:
I've found more nested classes that need to be annotated with `org.immutables.value.Generated`.

* `*.GsonAdapters*.*NamingFields`
* `*.GsonAdapters*.*TypeAdapter`
* `*.*Repository.Criteria`
* `*.*Repository.Finder`
* `*.*Repository.Modifier`
* `*.*Repository.Replacer`
* `*.*Repository.Serialization`
* `*.*Repository.*NamingFields`

It would be great to fix that because tools like JaCoCo interpreting it.

@elucash I'm hungry for a new release with these changes.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Gsons.generator
+++ b/value-processor/src/org/immutables/value/processor/Gsons.generator
@@ -133,6 +133,9 @@ public final class [typeAdaptersName] implements TypeAdapterFactory {
 [checkAttributes type]
 [for allAttributes = type.allMarshalingAttributes, t = type.typeAbstract, im = type.typeImmutable]
 
+[if type allowsClasspathAnnotation 'org.immutables.value.Generated']
+@org.immutables.value.Generated(from = "[type.typeAbstract.relativeRaw]", generator = "Gsons")
+[/if]
 @SuppressWarnings({"unchecked", "raw"}) // safe unchecked, types are verified in runtime
 private static class [type.name]TypeAdapter[type.generics] extends TypeAdapter<[t]> {
   [for a in allAttributes]
@@ -165,6 +168,9 @@ private static class [type.name]TypeAdapter[type.generics] extends TypeAdapter<[
 [if type.gsonTypeAdapters.fieldNamingStrategy andnot type.useConstructorOnly]
   [if allAttributes]
 
+  [if type allowsClasspathAnnotation 'org.immutables.value.Generated']
+  @org.immutables.value.Generated(from = "[type.typeAbstract.relativeRaw]", generator = "Gsons")
+  [/if]
   static class [type.name]NamingFields[type.generics] {
   [for a in allAttributes]
     public [a.type] [a.name];
--- a/value-processor/src/org/immutables/value/processor/Repositories.generator
+++ b/value-processor/src/org/immutables/value/processor/Repositories.generator
@@ -232,6 +232,9 @@ public Finder find(Criteria criteria) {
  * Configure exclusion and sort ordering for results using the family of {@code exclude*()} and {@code orderBy*()} attribute-specific methods.
  * @see [type.name]Repository#find(Criteria)
  */
+[if type allowsClasspathAnnotation 'org.immutables.value.Generated']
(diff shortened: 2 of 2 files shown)


value-processor/src/org/immutables/value/processor/Gsons.generator
```diff
@@ -133,6 +133,9 @@ public final class [typeAdaptersName] implements TypeAdapterFactory {
 [checkAttributes type]
 [for allAttributes = type.allMarshalingAttributes, t = type.typeAbstract, im = type.typeImmutable]
 
+[if type allowsClasspathAnnotation 'org.immutables.value.Generated']
+@org.immutables.value.Generated(from = "[type.typeAbstract.relativeRaw]", generator = "Gsons")
+[/if]
```
Stephan202: Elsewhere the following construct is used instead:
```
[if type allowsClasspathAnnotation 'javax.annotation.processing.Generated']
@javax.annotation.processing.Generated("org.immutables.processor.ProxyProcessor")
[else if type allowsClasspathAnnotation 'javax.annotation.Generated']
@javax.annotation.Generated("org.immutables.processor.ProxyProcessor")
[/if]
```

Grepping indicates that the different approaches are used with similar frequency:
```
$ git grep javax.annotation.Generated | wc -l
27
$ git grep javax.annotation.processing.Generated | wc -l
20
$ git grep org.immutables.value.Generated | wc -l
33 
```
@elucash is the plan to use `javax.annotation.processing.Generated` or `org.immutables.value.Generated` throughout the code base?

(IIRC `javax.annotation.processing.Generated` is pitched as the JDK 9+ replacement of `javax.annotation.Generated`.)

before: @Stephan202 as far as I understand it, we use `org.immutables.value.Generated` to tell tools like Jacoco during runtime what was generated. The Java types `javax.annotation.Generated` and `javax.annotation.Generated` have `@Retention(SOURCE)`.

Stephan202: @before good point about `@Retention`; I didn't consider that since I'm mostly interested in `@Generated` in the context of static analysis, such as using Error Prone.

I wonder: for tools that look for a fully qualified class name (rather than any annotation with simple name `@Generated`), would it make sense to add both annotations?

before: I don't know a tool yet that expects `@Generated` annotation on nested classes other then JaCoCo. Error Prone expects that everything is generated when the outer class is annotated with `javax.annotation.Generated`. I wouldn't spend time on this until we know a tool that needs it on nested classes.

elucash: Want to clarify:

* Source retained Generated (newer version `javax.annotation.processing.Generated` or older `javax.annotation.Generated`) is only used on a top level type, everything nested is considered generated without annotation. Newer takes precedence, obviously. They are used automatically if found on compilation classpath during annotation processing and can be disabled by whitelist override in styles or blacklisting via classpath fence (extension resource in META-INF).

```java
[if type allowsClasspathAnnotation 'javax.annotation.processing.Generated']
@javax.annotation.processing.Generated("org.immutables.processor.ProxyProcessor")
[else if type allowsClasspathAnnotation 'javax.annotation.Generated']
@javax.annotation.Generated("org.immutables.processor.ProxyProcessor")
[/if]
```

* Class retained `org.immutables.value.Generated` is covering each generated type, both top level and nested, to end up in each generated `*.class` file. It is always on the classpath (our annotation API jar has `Value.*` and `Generated` packaged together), but can be turned off by whitelist override in styles.

```java
[if type allowsClasspathAnnotation 'org.immutables.value.Generated']
@org.immutables.value.Generated(from = "[type.typeAbstract.relativeRaw]", generator = "Gsons")
[/if]
```

* If someone needs some runtime retained annotation, the option is to use (at least to try) annotation injection. (see `annotate` module)

elucash:
Looks good! Thank you!

Issue #804, opened 2018-07-23
Support Gradle Incremental Processing


khaidinh:
Gradle 4.7+ has a new feature to let annotation processors run incrementally. In order to make that work, Immutables has to tell Gradle how its annotation processor behave during execution.

For details see: 

https://docs.gradle.org/nightly/userguide/java_plugin.html#sec:incremental_annotation_processing

https://github.com/gradle/gradle/issues/5277

Supporting this feature, would greatly reduce build times using Gradle.

MartyIX:
@elucash Hi, could you please just give some pointers how to move this forward? If it is in line with your plans, of course.

I have put together some resources to understand what the issue is about:

* https://docs.gradle.org/nightly/userguide/java_plugin.html#sec:incremental_annotation_processing
* https://github.com/gradle/gradle/issues/5277
* https://github.com/JakeWharton/butterknife/pull/1230/files
* https://github.com/immutables/immutables/issues/804
* https://github.com/evernote/android-state/pull/49/files

elucash:
Thank you for the feature request! I think it would be really great to support this. I would definitely like such a feature if this would not require a lot of work.

Looking at examples I barely understand what exactly has to be done here (with respect to annotation processing code). Is there a way to externally tell / configure Annotation processor to be run in one of these sandbox/isolation modes and see if processor works before actually augmenting annotation processor's metadata?

+ _Immutables_ is already incremental capable, like when working inside of Eclipse on ECJ with auto-build on, so in theory we should be ok if incrementality enabled
+ Some features will not work or work slightly differently if put in Gradle isolation sandbox, depending how properly Gradle wrappers are coded (as both Javac and ECJ dev screwed up implementing those elements properly, little hope that Gradle will get it right early on)
+ PRs are welcome. The main problem here is not implementing any change, but ensuring things work properly and I definitely need help here. One of the testing strategies might be copying `value-fixture` content under Gradle build and see how much of edge cases compile.

[...]

PR #858, opened 2018-10-28
Allow opt-in to Gradle incremental processing support via -Aimmutable…

anuraaga:
…s.gradle.incremental

By providing a flag for opting in to incremental processing, users can try it out now and if there are any issues we'll be able to fix them (given the processor's fairly easy to reason about 1:n mapping of input to output, I guess no one should run into any in practice). Without the flag, there is no change in behavior for current users.

For #804 (not sure whether to consider it a fix since presumably it should be made default at some point).

### DIFF
--- a/value-processor/src/META-INF/gradle/incremental.annotation.processors
+++ b/value-processor/src/META-INF/gradle/incremental.annotation.processors
@@ -1,0 +1,2 @@
+org.immutables.processor.ProxyProcessor,dynamic
+org.immutables.value.processor.Processor,dynamic
--- a/value-processor/src/org/immutables/value/processor/Processor.java
+++ b/value-processor/src/org/immutables/value/processor/Processor.java
@@ -16,6 +16,7 @@
 package org.immutables.value.processor;
 
 import com.google.common.collect.FluentIterable;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
 import java.util.Set;
 import javax.annotation.processing.SupportedAnnotationTypes;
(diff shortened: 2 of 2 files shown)


elucash:
@anuraaga Thank you for the pull request! To be precise, there are many features that will work properly in isolating mode (as incremental compilation works in Eclipse right now), while some of the features will not work (generating single Gson adapter across the package). Assuming that those features are not the key features and can be avoided, the usage of `-Aimmutables.gradle.incremental` should/will be documented with a caution message. I would be cool to issue some compilation warnings (or even errors) when isolating mode is enabled, it just requires some analysis of which specific features are affected.

oehme:
In that case, maybe you should offer the user both options? Because the single Gson adapter will work just fine in "aggregating" mode.

anuraaga:
I personally think that Gradle should allow users to provide a `Processor` -> incremental mode `Map` if they'd like (when I tried, adding my own `META-INF` to the `annotationProcessor` classpath didn't work, I presume the metadata has to be in the processor's JAR to be picked up). Annotation processors can provide a default (which applies to 100%, or 90%, or whatever the author wants to target, it's their call) if they want, while a Gradle API providing fine-grained customization would apply to all processors out there without code change and complexity. But I don't feel too strongly, just my thoughts after writing this PR.

oehme:
> apply to all processors out there without code change and complexity

This mostly wouldn't work or would be highly unsafe - Only the processor author knows how their processor works and what kind of incrementalism they can support.

elucash:
I can speculate we cannot determine these features upfront as they are not global, but depends on the placing on annotations (so we cannot switch to aggregating mode dynamically), but once we know that we use them in isolating mode, then we can for instance, issue warnings, but even as the annotation processor author I don't know all the affected features exactly. This new `-A` flag could be advertised as "use if it works for you" optimization.

Issue #732, opened 2017-12-11
Provide an option to clone/copy a builder


billccn:
Currently there's no generic way to copy a builder especially those with validations.

I write a lot of test setup code that manipulates builders instead of creating actual domain objects so test code can be easily reused. A lot of these need the ability to copy builders.

Since the Immutable framework knows exactly how to copy each field, it shouldn't be too difficult to implement a method that copies everything and skips any validations

elucash:
Thank you for reporting this. Currently, you can use builders' from method to fill builder with values of an instance: some exemplary instances can be created as prototypes so other instance's builders copy those values and override them. Builders can be used to create instances of one set of attributes, then reconfigured to create instances another set of attributes. Is there any particular pressure to copy/clone builders themselves? If I understand correctly, some incomplete builders have to be cloned to be completed and then built in different ways. I'll see how easy/difficult it will be to implement this. In a meantime, I would recommend taking a closer look at [`@Value.Modifiable`](http://immutables.github.io/immutable.html#modifiable-classes) to create more powerful buffer object than builders (no guarantees that these can support your use case, but still). Another workaround can be using 2 kinds of objects: final objects and prototypes: not validated, maybe with just a subset of fields shared by supertype:

```java
interface Final extends CommonFields { int a(); String b() }
@Immutable interface Proto extends CommonFields {}
@Immutable interface Final extends CommonFields { boolean c(); default @Check void check() {...} }

proto = ImmutableProto.builder().a(1).b("b").build(); // unvalidated subset
ImmutableFinal.builder().from(proto).c(true).build(); // fully built instance
```

PR #872, opened 2018-12-03
buildFromAttribute: treat all non-primitive attributes as potentially nullable

stevenschlansker:
We generate Builder `from` methods for super-interfaces of the type we are generating.
There is no guarantee that the super-interface has the same characteristics, in particular, it may have a "wider" type that includes null.

buildFromAttribute uses the local type's idea of whether an attribute is null.

This means if you have a sub-interface that specializes an attribute to `@Nonnull`, it may incorrectly assume that the super-type *also* declared the attribute as `@Nonnull`, when in fact it is legitimate for it to be `@Nullable` (since it is a wider type).

This proposed change treats all reference attributes as potentially nullable while building.  There should (🤞) be no performance impact as the getters will be inlined and the JVM is very good at optimizing impossible null checks.

Please let me know if you see a better way to fix this, or an error with my logic.  Thanks!

### DIFF
--- a/value-fixture/src/org/immutables/fixture/nullable/BaseNullableAttribute.java
+++ b/value-fixture/src/org/immutables/fixture/nullable/BaseNullableAttribute.java
@@ -1,0 +1,22 @@
+package org.immutables.fixture.nullable;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import org.immutables.value.Value;
+
+public interface BaseNullableAttribute {
+    @Nullable
+    String attribute();
+
+    @Value.Immutable
+    public interface NullableAttribute extends BaseNullableAttribute {
+    }
+
+    @Value.Immutable
+    public interface NonnullAttribute extends BaseNullableAttribute {
+        @Override
+        @Nonnull
+        String attribute();
+    }
+}
--- a/value-fixture/test/org/immutables/fixture/nullable/NullableAttributesTest.java
+++ b/value-fixture/test/org/immutables/fixture/nullable/NullableAttributesTest.java
@@ -257,4 +257,13 @@ public class NullableAttributesTest {
     check(r.i1()).isNull();
     check(r.l2()).isNull();
   }
+
+  @Test
+  public void sometimesNullableFrom() {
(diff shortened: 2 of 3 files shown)


stevenschlansker:
Any thoughts on this fix?  We're relying on it in our projects and would like to make sure it makes it into `master` for the next release.  Thanks!

stevenschlansker:
Ah, that's unfortunate it doesn't fit in the model.  Let me explain my motivation more clearly:

We want to generate a `Foo` type, and a `FooPatch` type to represent partial updates. (Roughly the same pattern you outline here: https://github.com/immutables/immutables/issues/732#issuecomment-350961107 )

So we define:

```java
interface FooFields {
    @Nullable
    String field1();
    @Nullable
    String field2();
}

@Value.Immutable
interface Foo extends FooFields {
    @Nonnull
    @Override
    String field1();

    @Nonnull
    @Override
    String field2();
}

@Value.Immutable
interface FooPatch extends FooFields {}

// No problem with the first store
Foo value =  ImmutableFoo.builder()
    .field1("a")
    .field2("b")
    .build();

// But when you later try to patch it...
Foo updatedValue =
    ImmutableFoo.builder()
        .from(value)
        .from(FooPatch.builder().field2("c").build()) // boom, we try to set null 'field1'
        .build();
```

and that's how we find ourselves in this situation.

I hope we can find a way to make this work.  Thanks.

elucash:
Definitely I see a rationale, what I'm trying to do is to fit it
conceptually, for example, if we clearly see (inspected) that supertype
method is nullable and immutable one is not, then we add null check because
it would be necessary and sound. I'm against implementing this in
"defensive" coding style where everything can be null and we mask things
with noops

elucash:
@stevenschlansker thank you again for raising this and contributing PR. I've integrated the part which would check if we have a non-nullable attribute which is also inherited from supertype where it's declared as nullable, so we insert null-check in `from` method. As I mentioned, the approach where we expect any non-primitive value to be nullable would not play well with our strong preference to fail-fast on null than silently ignored. But marking nullable in subtype and overriding with non-null should give this effect as opt-in rather than universal behavior. thank you!

PR #881, opened 2018-12-13
Stand-alone Criteria API to query different data-sources (initial draft)

asereda-gs:
First draft of Criteria API. Please don't merge it yet.

Generate criteria class based on existing model and evaluate it using visitor pattern.

Right now only reflection based evaluator is provided. 

### Missing functionality / Remaining Questions : 
1. (sub-)Criteria on Iterables. Example 
    ```java
    user.friends.anyMatch(FriendCriteria.create().age().greaterThan(22));
    user.friends.noneMatch(FriendCriteria.create().age().greaterThan(22))
    user.friends.allMatch(FriendCriteria.create().age().greaterThan(22));
   
    // something similar for optionals
    ```
2. Combining criterias (using `AND`s / `OR`s)
3. Nested criterias
4. Projections
5. Aggregations
5. Comparison between fields: `field1 = field2` 
6. DSL is not restrictive enough.
   ```java
   // one can write something like this
   criteria.or().or().or();
   criteria.age().greaterThan(111).or().or(); // or this
   ```
7. How are we better / worse than [QueryDSL](http://www.querydsl.com/) ?
8. Allow user to define (inject) custom (native) queries like [geo spacial](https://docs.mongodb.com/manual/geospatial-queries/)

### DIFF
--- a/criteria/README.md
+++ b/criteria/README.md
@@ -1,0 +1,26 @@
+This module generates Criteria classes based on immutable model.
+
+Generated classes allow type-safe query building which can be evaluated on destination data-source.
+
+Conversion between criteria and native query is happening at runtime using visitor pattern.
+
+### Missing functionality / Remaining Questions : 
+1. (sub-)Criteria on Iterables. Example 
+    ```java
+    user.friends.anyMatch(FriendCriteria.create().age().greaterThan(22));
+    user.friends.noneMatch(FriendCriteria.create().age().greaterThan(22))
+    user.friends.allMatch(FriendCriteria.create().age().greaterThan(22));
+   
+    // something similar for optionals
+    ```
+2. Combining criterias (using `AND`s / `OR`s)
+3. Nested criterias
+4. Projections
+5. Aggregations
+6. DSL is not restrictive enough.
+   ```java
+   // one can write something like this
+   criteria.or().or().or();
+   criteria.age().greaterThan(111).or().or(); // or this
+   ```
+7. How are we better than [QueryDSL](http://www.querydsl.com/) ?
--- a/criteria/pom.xml
+++ b/criteria/pom.xml
@@ -1,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- 
+   Copyright 2014-2018 Immutables Authors and Contributors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>immutables</artifactId>
+    <groupId>org.immutables</groupId>
+    <version>2.7.6-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>criteria</artifactId>
+  <name>${project.groupId}.${project.artifactId}</name>
+  <description>
+    Annotation and runtime support to generate criteria classes based on immutables model
+  </description>
+
+  <dependencies>
+    <dependency>
+      <!-- Guava and jsr 305 annotations are used as required utility.-->
+      <groupId>org.immutables.dependency</groupId>
+      <artifactId>utility</artifactId>
+      <version>${project.version}</version>
+      <type>pom</type>
+    </dependency>
+
+    <dependency>
+      <groupId>org.immutables</groupId>
+      <artifactId>value</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <!-- compile only meta-inf/services generator -->
+      <groupId>org.immutables</groupId>
+      <artifactId>metainf</artifactId>
+      <version>${project.version}</version>
+      <scope>provided</scope>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.immutables</groupId>
+      <artifactId>testing</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+</project>
--- a/criteria/src/org/immutables/criteria/Criteria.java
+++ b/criteria/src/org/immutables/criteria/Criteria.java
@@ -1,0 +1,35 @@
+/*
+   Copyright 2013-2018 Immutables Authors and Contributors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ */
+
+package org.immutables.criteria;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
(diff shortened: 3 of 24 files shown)


asereda-gs:
Probably we should make several maven (sub-)modules ?
- `criteria/core` (core functionality and runtime dependencies)
- `criteria/mongo`
- `criteria/elastic`
- `criteria/jdbc` 
 
The query / fetch API can be based on [Reactive Streams](http://www.reactive-streams.org/) ? For instance criteria can be applied on the top of [change streams](https://docs.mongodb.com/manual/changeStreams/).

Besides criteria should one also generate Repositories for each data-source (what has been done for mongo) ?

asereda-gs:
Hi @elucash . I'm planning to add more functionality. Are you OK with the approach taken so far ?

elucash:
@asereda-gs I'm looking into it, have some ideas to try. There are other 1 high priority item left to make a release today, then I'm fully into this

elucash:
released 2.7.4, fetching this PR to tinker with it locally

asereda-gs:
@elucash pls let me know if you have feedback on this initial PR ?

elucash:
@asereda-gs my apologies, every time I'm going to dedicate some time to look into it, something else takes away all energy (

I definitely like what I see here, have some thought about it

1. I really liked (at some point in time) idea that we can get attribute's criteria expressed as fields, which would create nice syntactic contrast with the criteria methods `age.greaterThan(4)` as opposed to `age().greaterThan()` . I don't want it to a contentios point, if there are good arguments for all-methods we can go with it. Just thinking that "blind consistency" should not be a factor here (as well as formatting stuff as you pointed out)

2. Franky I don't know if we should improve this, just thinking aloud, `FriendCriteria.create()` is a bit non dsl-ish, which can be though of as a factory equivalent of `new FriendCriteria()`. For me `create` suggests the we will always create a new object, but empty immutable criteria can be a singleton. When it was obtained from the repo object (or class), it could be streamlined accessor and then dsl-ish var name `where` etc. `Criteria where = posts.criteria(); where.content("!!!").ratingsNonEmpty()`. This presents some chance to play with that factory name as `create` is too bland and carries no additional info (being equiv of `new`). However adding too much of gimmick to is also, probably, not the best idea. `FriendsCriteria.where()`... `FriendsCriteria.start()`

3. Need to think about extending criterias, possibly by combining with existing criteria. For example let's say we add some type like `Money`, but at the same time it's also `Comparable`. Here, first idea is to use interfaces (&default methods) to create such combined criterias (akin to [that](https://github.com/immutables/io/blob/master/src/io/immutables/that/That.java) and [that](https://github.com/immutables/io/blob/master/src/io/immutables/that/Assert.java) ). Also in light of this, the structure with visitors (contraint collectors) might be no longer a good idea, and looking back, I think it was quite fancy but not that good solution.

asereda-gs:
@elucash thanks for your reply. Some answers to your feedback : 

1. I'm not fundamentally against having fields in criteria (as opposed to methods). We can agree on `age.greaterThan(4)` (instead of  `age().greaterThan()`). 

2. The reason there is factory method `FriendCriteria.create()` is because I wanted to make criteria independent of repository implementation. Same criteria should be applicable to mongo, elastic , JDBC repositories. To some extent I'm fine with `new FriendCriteria()` but, ideally, would like to keep  constructor private (for more control). Perhaps expose static field `FriendsCriteria.DEFAULT / INSTANCE` (not very nice)? 

3. It is true that with default methods (Java 8) there is more flexibility, but don't we have to stay compatible with Java 7 (due to immutables) ? Will immutables library migrate to Java 8 soon ? 
To me, constraints API was just a way to "serialize" expression (similar to linked list) and visitor pattern was used to evaluate it. Both are private APIs and are not exposed to the user. Being able to serialize criteria(s) is an appreciated feature. Do you mind sharing some code snippets of how  you see ideas from _That_ library included in criteria ? 

I'm happy to implement those ideas but would like to agree with you on API and generated code first.

elucash:
in 2) I'm not questioning the overall composition with the factory method (which could hide both constant or new instance creation), but mery opportunistically attempt to come with a fancier name (maybe add so dsl-ishness to it)

Let me know if you have specific plans to use java7, but otherwise j8 is fine. We maintain compatibility for the core (which we will drop in hypotetical Immutables3) so if people already use immutables 2.x they can still upgrade. But some parts of our functionality is already Java8 only, so if we create addition generator we can freely choose to go J8+.

Visitors, in my view, cannot give us good extensibility. Alternative is some other form of handlers to allow for the same functional effect yet do not constain to a "seal" set of methods. Fully agree that some examples / snippets would be a good way to show it

asereda-gs:
Re java 7. We're not using it and happy to see java8 dependency. Does it mean criteria (maven) module would be on 8 while _core_ on 7 ? 

## Visitor 
For visitors you don't have to "seal" all the methods. One can have generic enough expressions abstracted. At the end you still need to "evaluate" constructed criteria and I don't know how to perform it generically without listener / visitor pattern. 

### Example
```java
// Allows visiting criteria expressions
interface Visitable {
    <R, V extends Visitor<R>> R accept(V visitor);
}

interface Visitor<R> {
   R visit(Expression<?> expr);
}

interface Expression<C> {

   Expression<?> argAt(int index);

   List<Expression<?>> args();
   
   // defines generic operation like EQUALS, IN, GREATER_THAN, etc. 
   Operator operator();
}

// marker interface ? 
interface Operator {}

// default (or predefined) operations
enum Operators implements Operator {
  EQUALS, 
  NOT_EQUALS,

  IN,
 NOT_IN, 
  GREATER_THAN, 
  // ....
}
```

elucash:
fixture module is already on java 8, the processor/template would be on 7,
but new module api and support classes will be 8, should work fine.

will be creating some examples of non-visitor etc

asereda-gs:
@elucash is there anything I can help with to speedup this PR ?

elucash:
Really like the changes, will skim over once again and merge it. It's still "experimental" so should be fine. At the same time I will try looking for "extensibility" models if there are any good ones

asereda-gs:
Thank you, Eugene. Can you please also give an example of what you mean by "non-visitor API" ? 

BTW I would like to make some modifications to existing `mongo` repositories. Will you have time to review those ?

asereda-gs:
@elucash I have fixed all build errors. You're welcome to merge it.

However, please note that criteria requires more work (see `Remaining Questions`) in PR description. I would like to make those contributions over next couple of weeks. Will you have time to review them ? When / How would you like those PRs be submitted ?

asereda-gs:
@elucash do you agree to create more modules under `criteria/` for example mongo / elastic "adapters" :  

1. `criteria/core` shared code ? 
2. `criteria/mongo` for mongo ?
3. `criteria/elastic` for elastic ? 

Should repository pattern be based on [Reactive Streams](http://www.reactive-streams.org/) instead of guava's / JDK8 futures ?

elucash:
modules are ok, just few notes about how I choose names there
so the first rule of thumb is to try to avoid meaningless cliché names like  `core`, `kernel`, `engine`, `util`, `common` etc (I do realize that sometimes it's unavoidable, but still). If we're talking about packages then it can be `whatever.criteria`, `whatever.criteria.mongo`, `whatever.criteria.elastics` . If it's about directory structure, there all should be under one root, then either a) avoid entropy and just repeat `criteria/criteria`, then `criteria/mongo`, `criteria/elastic` or b) use more cliché such as "shared" or "common" which is in my opinion more honest/precise than "core", `criteria/shared` (also a) is still a better option in many cases)

We can use reactive streams, especially if we're talking about related `java.*` classes. On the other hand, it would be beneficial to have pluggability/adapters for say JDK8 futures (as batteries included) or third-party ones (would not be included, probably, but not so difficult to code)

asereda-gs:
Hi @elucash,

I'm looking for a solution (advice) on how to deal with some legacy code. We have an in-house (old) code-gen which generates POJOs from [Json Schema](https://json-schema.org/) (along with immutables). There is about 30/70 distribution of code generated by in-house and immutables libraries respectively. I would love to migrate legacy code to immutables but existing code depends on mutability of the classes.

In-house code-gen looks similar to the class below : 
```java
// can also extend and/or implement classes / interface
public class Model {
    // plus other things to detect if class was mutated

    private String foo;
  
    // public no-args constructor 
    public Model() {}

    public void setFoo(String foo) {
         this.foo = foo;
    }
 
    public Model withFoo(String foo) {
        this.foo = foo;
        return this;
    }

     public String getFoo() {
         return this.foo;
     }
  
}
```

Ideally my goal would be to generate immutables Criteria based on old model. Do you think there is a way for criteria module to read external model and generate only criteria classes (not others) ? Or, maybe a bigger undertaking, a hook in immutables to generate this "custom" POJO ? 

Many thanks for your help.

elucash:
Hi Andrei! In theory, this can work and there are some precedents of something similar, first example is builder generated for pojos (based on constructor parameters) or for static factory methods (based on static method parameters). in both cases those goes thru the same "legacy/core" ValueType/ValueAttribute model machinery, yet requires a number of special casing in templates. If you're interested in following this path, the jorney starts in ~~the Castle of Blackmarsh~~ `ValueTypeComposer` where the model is constructed from different kind of inputs (abstract classes / pojo constructors, static factory methods etc). I dunno if it's a suitable choice, just the first thing that popped from the top of my head.

PR #884, opened 2018-12-18
Fix parent package traversal in JDK 9+

### LINKED ISSUE #736: Why declared immuables library style in package-info.java doesn't apply for subpackages in java 9 project?
Please see https://github.com/originalrusyn/java9_immutables for more details

So, the following config works correctly with sourceCompatibility = 1.8

```
p.package-info.java

@Value.Style(
    defaults = @Value.Immutable(copy = false),
    strictBuilder = true,
    overshadowImplementation = true,
    implementationNestedInBuilder = true
)
package p;

```

build.gradle

```
buildscript {
    ext {
        immutablesVersion = '2.5.6'
    }
    repositories {
        mavenCentral()
        maven { url "https://plugins.gradle.org/m2/" }
    }
    dependencies {
        classpath "net.ltgt.gradle:gradle-apt-plugin:0.13"
    }
}

group 'java9'
version '1.0-SNAPSHOT'

apply plugin: 'java'
apply plugin: 'net.ltgt.apt'

sourceCompatibility = 1.9

repositories {
    mavenCentral()
}

sourceSets {
    main
    generated
    test
}

compileJava {
    generatedSourcesDestinationDir = "$projectDir/src/generated/java"
}

task wrapper(type: Wrapper) {
    gradleVersion = '4.4'
}

dependencies {
    apt "org.immutables:value:${immutablesVersion}"

    compileOnly "org.immutables:value:${immutablesVersion}:annotations"

    testCompile group: 'junit', name: 'junit', version: '4.12'
}
```

p.b.B.java

```
package p.b;

@Value.Immutable
public interface B {

    String getText();
}
```

Why does it produce p.b.ImmutableB.java rather than p.b.BBuilder.java for sourceCompatibility=1.9?

Stephan202:
Fixes #736.

Summary of our situation and how I came to this proposal: We have a small Maven module with a `com/company/package-info.java` file containing a company-wide `@Value.Style` configuration. Other Maven modules declare `@Value.Immutable` types in `com/company/foo/bar/baz` packages. With JDK 8 this works fine; with JDK 9-11 it works only after applying this patch. My understanding of  the problem is that with JDK 9+, Immutables only finds `com/company/foo/bar` as a parent of `com/company/foo/bar/baz` if the former contains a `package-info.java` file. The patch proposed here doesn't bail out when the parent is not found, but instead keeps traversing to the root, since packages higher up _may_ contain a `package-info.java` file.

@elucash I tested this change on a private project using JDK 9, 10 and 11. I wanted to add a test case, but since the project currently doesn't build with JDK 9 anyway (there are test failures in `org.immutables.gson`), that seems moot. Let me know if you disagree :)

### DIFF
--- a/value-processor/src/org/immutables/value/processor/meta/Proto.java
+++ b/value-processor/src/org/immutables/value/processor/meta/Proto.java
@@ -723,7 +723,7 @@ public class Proto {
     @Value.Lazy
     Optional<DeclaringPackage> namedParentPackage() {
       String parentPackageName = SourceNames.parentPackageName(element());
-      if (!parentPackageName.isEmpty()) {
+      while (!parentPackageName.isEmpty()) {
         @Nullable PackageElement parentPackage =
             environment().processing()
                 .getElementUtils()
@@ -737,6 +737,11 @@ public class Proto {
                   .element(parentPackage)
                   .build()));
         }
+
+        // With JDK 9+ package elements are only returned for packages with a
+        // `package-info.class` file. So although the parent may not be found,
+        // there may be "ancestor packages" further up the hierarchy.
+        parentPackageName = SourceNames.parentPackageName(parentPackageName);
       }
       return Optional.absent();
     }
--- a/value-processor/src/org/immutables/value/processor/meta/SourceNames.java
+++ b/value-processor/src/org/immutables/value/processor/meta/SourceNames.java
@@ -46,7 +46,10 @@ final class SourceNames {
   }
 
   static String parentPackageName(PackageElement element) {
-    String qualifiedName = element.getQualifiedName().toString();
+    return parentPackageName(element.getQualifiedName().toString());
+  }
+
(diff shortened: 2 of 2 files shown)


elucash:
@Stephan202 Thanks a lot! I never had time to look into this deeply, but that's really good we have it fixed if we nailed it down. I'll merge this for the ease of trying it out for the other folks.

Stephan202:
@elucash thanks for merging this so rapidly! Do you plan to cut a new release soon? I'm asking to assess whether I should wait a bit or temporarily switch over our internal code base to a fork that includes this patch. (No pressure, either answer is totally fine.)

elucash:
@Stephan202 I thought of releasing in 1-2 days, depending on progress. Dunno if this is soon enough for your needs ;)  Thanks!

Stephan202:
Works for me, thanks! :smile:

PR #893, opened 2018-12-31
Use Guava static methods for better forward compatibilty.

stevestorey:
Guava 26 removed the static constants being used by Immutables, which can make
it very difficult to use Immutables in a project with a newer Guava version.
Version 19 (which happily is already in use here) provided static methods to
replace the constants, so switch to using them.

### DIFF
--- a/generator-processor/src/org/immutables/generator/processor/Trees.java
+++ b/generator-processor/src/org/immutables/generator/processor/Trees.java
@@ -334,7 +334,7 @@ public class Trees {
     }
 
     public boolean isWhitespace() {
-      return CharMatcher.WHITESPACE.matchesAllOf(value());
+      return CharMatcher.whitespace().matchesAllOf(value());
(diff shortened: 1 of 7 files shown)


elucash:
for a large portion of the codebase we maintain JDK 1.7 compatibility, Guava after its version 20 moved to 1.8, so we may postpone such updated until safe, so this PR will hang for a while. As you're immediate need was addressed by using `value` artifact, I assume the update is not critical at this moment. Thank you!

stevestorey:
This PR is perfectly compatible with the existing Guava 19 (I'm not upgrading it) and (though I've not tested it) should be fine with JDK 1.7 as well. But as you point out, actually following the instructions :) works, and this update isn't critical.

elucash:
ah, overlooked that aspect. Then I'll verify it and merge, thanks!

PR #899, opened 2019-01-04
Don't write file if content is identical

henrikno:
I noticed we were recompiling some modules even without doing any changes, and traced it back to files generated by immutables that changed the mtime of the files. This PR attempts to speed up incremental compiles by not writing the file if the content is already the same.

### DIFF
--- a/generator/src/org/immutables/generator/Output.java
+++ b/generator/src/org/immutables/generator/Output.java
@@ -310,12 +310,14 @@ public final class Output {
     void complete() {
       CharSequence sourceCode = extractSourceCode();
       try {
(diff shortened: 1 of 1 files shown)


elucash:
@henrikno thank you for the PR! very clever solution. Sorry for the delay, I've seen this but only now got a chance to look into. So while I definitely see the benefits, the other side of this is that there should be an impact for any other usage (on this later) as reading the whole files and running them thru equals might have some performance impact that I'm yet to estimate. I'm interested to know the specifics of incremental compilation (was it just javac, or maven, gradle or something else). Is it possible to do this only when incremental compilation is executed? (I know this might be hard to answer)

PR #910, opened 2019-02-04
add Builder.from() support for direct ancestors with just one type parameter

tpetr:
We use the [egg pattern](https://github.com/HubSpot/hubspot-immutables#egg-pattern) heavily at my company and this PR addresses an issue we've faced where `Builder.from()`'s are not generated for supertypes when all types have the same type parameter. To see what I mean, check out this test I added (https://github.com/immutables/immutables/compare/master...tpetr:generic-inherit?expand=1#diff-f5bc5bf0a9b538e82ec484267fceffdd) -- I need to be able to pass a `ImmutableGenericChild` to `ImmutableGenericAdult.Builder.from()`.

This PR adds support for direct supertypes where both classes have just one type parameter with matching bounds and names. I wasn't able to safely reason about anything more complicated than that 😓 

Let me know if this all makes sense or if there's anything I can do to make this more merge-able. Thanks!

### DIFF
--- a/value-fixture/src/org/immutables/fixture/generics/GenericCore.java
+++ b/value-fixture/src/org/immutables/fixture/generics/GenericCore.java
@@ -1,0 +1,17 @@
+package org.immutables.fixture.generics;
+
+import org.immutables.value.Value.Immutable;
+
+public interface GenericCore {
+  interface Core<T extends CharSequence> {
+    T getObject();
+  }
+
+  @Immutable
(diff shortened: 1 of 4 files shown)


daicoden:
I had the same problem, as a work around if you override the method in the child interface, you will get the proper `from` generated.

elucash:
thank you! seems like quite narrow focused addition which can be useful

PR #916, opened 2019-02-24
ignore methods with parameters when comparing with super type

scrozon:
if a class defines a method with the same name as a "getter" method but with parameters, and that method is declared before the getter, it will be used to compare the return type against the getter method defined in the parent class, if they don't match, the attribute will be ignored from the "from" builder.

although it's probably not a common use case to have methods with parameters named like a getter, only methods without parameters should be considered anyway. 

I couldn't find unit tests for this, so didn't edit any, but happy to add test cases if there are tests for this in another module.


elucash:
looks like a good idea, thank you!

PR #927, opened 2019-03-21
split criteria into maven sub-modules: common, mongodb and elasticsearch

asereda-gs:
Current PR separates existing criteria maven module into the following:

1. `criteria/common` (do you prefer `criteria/criteria` or `criteria/shared`?)
2. `criteria/mongodb` (do you prefer`criteria/mongo` ?)
3. `criteria/elasticsearch` (do you prefer`criteria/elastic`?)

Migrate existing criteria code into `criteria/common`

Let me know if you agree with module / package naming


elucash:
my preference so far, not insisting, however

criteria/common
criteria/mongo
criteria/elastic

elucash:
ok, with elastic can be confusion (elastic cache, elasticsearch etc), having `elastics` is kinda criptic, so `elasticsearch` should be just fine, sorry for back-and-forth thoughts ))

PR #930, opened 2019-03-23
Mongo Adapter. Replace Fongo by Mongo Java Server (embedded mongo for tests)

### LINKED ISSUE #929: Mongo Adapter. Replace Fongo by Mongo Java Server (embedded mongo for tests)
[Mongo Java Server](https://github.com/bwaldvogel/mongo-java-server) looks like an actively maintained alternative to [Fongo](https://github.com/fakemongo/fongo).

With this change upgrade of mongo driver is possible.

See #831 and #887

asereda-gs:
Traditionally we have used [Fongo](https://github.com/fakemongo/fongo) as our fake mongo server. Unfortunately,
development on that project has staled and it is not compatible with latest mongo driver versions (3.6+).

[Mongo Java Server](https://github.com/bwaldvogel/mongo-java-server) is more active and passed all integration tests.

Closes #929

### DIFF
--- a/mongo/pom.xml
+++ b/mongo/pom.xml
@@ -72,12 +72,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>com.github.fakemongo</groupId>
-      <artifactId>fongo</artifactId>
-      <version>2.1.0</version>
+      <groupId>de.bwaldvogel</groupId>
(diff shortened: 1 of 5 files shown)


elucash:
👍 I like this approach way better! tnx!

Issue #934, opened 2019-03-28
Gson serialization generates incorrect code for polymorphic mapping


DipaanB:
I am facing a compilation error in the auto-generated Repository for serializing a polymorphic type for writing to MongoDB.

It fails to compile with the following errors:
gson cannot be resolved to a variable, DeptRepository.java, line 491
TypeToken cannot be resolved, DeptRepository.java, line 493
TypeToken cannot be resolved, DeptRepository.java, line 494

Here is the relevant section of the incorrectly generated code:
```
this.personsEncoder = ExpectedSubtypesAdapter.create(gson,
          PolymorphicMappingTest.Person.class,
          TypeToken.get(Student.class),
          TypeToken.get(Teacher.class));
```
Here is the relevant snippet of the source code (_Dept contains a polymorphic list of Person objects. The Person objects will be instances of either Student or Teacher):
```
    @Immutable
    @TypeAdapters
    @Repository
    public static abstract class _Dept {
        abstract List<Person> persons();
    }

    @ExpectedSubtypes({Student.class, Teacher.class})
    public static abstract class Person {
        abstract String name();
    }

    @Immutable
    public static abstract class _Student extends Person {
        abstract int fees();
    }

    @Immutable
    public static abstract class _Teacher extends Person {
        abstract int salary();
    }
```
I am attaching a junit5 test to illustrate and recreate the problem. In my test, the generated immutable classes will have underscore removed from the name. I am attaching the full generated code for the Repository as well.

[code.zip](https://github.com/immutables/immutables/files/3016480/code.zip)

The source code is PolymorphicMappingTest.java.
The generated code is DeptRepository.java.

Versions I am using:
com.google.code.gson:gson:2.8.5
org.immutables:mongo:2.7.5
org.immutables:gson:2.7.5
org.mongodb:mongo-java-driver:3.8.1

elucash:
@DipaanB thank you for reporting this!
@asereda-gs can you please take a quick look if this is a leftover of gson to CodecRegistry transition,
the repository template generates following code

```java
    Serialization(CodecRegistry registry, RepositorySetup.FieldNamingStrategy fieldNamingStrategy) {
      this.registry = registry;
      this.personsEncoder = ExpectedSubtypesAdapter.create(gson,
          PolymorphicMappingTest.Person.class,
          TypeToken.get(Student.class),
          TypeToken.get(Teacher.class));
      this.personsName = translateName(fieldNamingStrategy, "persons");
    }
```

[...]

PR #935, opened 2019-04-03
Explicitly fail for repositories with Gson polymorphic mapping (@ExpectedSubtypes)

asereda-gs:
Since migration to [mongo codecs](http://mongodb.github.io/mongo-java-driver/3.10/bson/codecs/) a case of polymorphic serialization was missed (gson serialization). This seems to be an edge case and it is safer to fail fast during repository code generation until a proper solution is in place.

For more information see #934


elucash:
Thank you for completing it!

PR #938, opened 2019-04-03
Use Disjunctive Normal Form (DNF) with criteria (allows chaining of expressions).

asereda-gs:
Introduce `DnfExpression` which keeps track of conjuction / disjunction (`AND / OR`) as separate lists.
This gives a shallow expression tree.

### DIFF
--- a/criteria/common/src/org/immutables/criteria/constraints/DebugExpressionVisitor.java
+++ b/criteria/common/src/org/immutables/criteria/constraints/DebugExpressionVisitor.java
@@ -1,0 +1,46 @@
+package org.immutables.criteria.constraints;
+
+import java.io.PrintWriter;
+import java.util.Collections;
+import java.util.Objects;
+
+/**
+ * Used to output expression tree as string. Useful for debugging expressions.
+ */
+public class DebugExpressionVisitor<Void> implements ExpressionVisitor<Void> {
+
+  private final PrintWriter writer;
+
+  private int depth;
+
+  public DebugExpressionVisitor(PrintWriter writer) {
+    this.writer = Objects.requireNonNull(writer, "writer");
+  }
+
+  @Override
+  public Void visit(Call<?> call) {
+    writer.println();
+    writer.print(String.join("", Collections.nCopies(depth * 2, " ")));
+    writer.print("call op=" + call.getOperator());
+    for (Expression<?> expr: call.getArguments()) {
+      depth++;
+      expr.accept(this);
+      depth--;
+    }
+    return null;
+  }
+
+  @Override
+  public Void visit(Literal<?> literal) {
+    writer.print(" literal=");
+    writer.print(literal.value());
+    return null;
+  }
+
+  @Override
+  public Void visit(Path<?> path) {
+    writer.print(" path=");
+    writer.print(path.path());
+    return null;
+  }
+}
--- a/criteria/common/src/org/immutables/criteria/constraints/EmptyExpression.java
+++ b/criteria/common/src/org/immutables/criteria/constraints/EmptyExpression.java
@@ -1,0 +1,20 @@
+package org.immutables.criteria.constraints;
+
+import javax.annotation.Nullable;
+
+/**
+ * Sentinel for a null expression.
+ */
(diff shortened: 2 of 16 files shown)


criteria/common/src/org/immutables/criteria/constraints/DebugExpressionVisitor.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Used to output expression tree as string. Useful for debugging expressions.
+ */
+public class DebugExpressionVisitor<Void> implements ExpressionVisitor<Void> {
+
+  private final PrintWriter writer;
+
+  private int depth;
+
+  public DebugExpressionVisitor(PrintWriter writer) {
```
elucash: there's an idea to generalize this to `Appendable` but the thing I don't like about `Appendable` is checked IO exception on every append, so probably not a good idea also. Maybe `Writer` could be considered as `StringWriter` can be used. Anyway, this is not really important

criteria/common/src/org/immutables/criteria/constraints/EmptyExpression.java
```diff
@@ -0,0 +1,20 @@
...
+final class EmptyExpression<T> implements Expression<T> {
+
+  static final EmptyExpression INSTANCE = new EmptyExpression();
+
+  private EmptyExpression() {}
+
+  @Nullable
+  @Override
+  public <R> R accept(ExpressionVisitor<R> visitor) {
+    throw new UnsupportedOperationException("Can't visit empty expression");
```
elucash: do nothing could be ok for an "empty". But I wonder the nature of it, probably, I should think of it as something more like bottom-type, "nonexpression" rather than "unit" / "empty" product.

criteria/common/test/org/immutables/criteria/ReflectionTest.java
```diff
@@ -23,16 +27,27 @@ public void reflection() {
...
     check(!evaluate(crit.lastName.isPresent(), person));
     check(evaluate(crit.lastName.isPresent(), person.withLastName("Smith")));
-    check(!evaluate(crit.lastName.isEmpty(), person.withLastName("Smith")));
+    check(!evaluate(crit.lastName.isAbsent(), person.withLastName("Smith")));
   }
 
+  @Test
+  public void debug() {
+    PersonCriteria crit = PersonCriteria.create().lastName.isAbsent()
+            .or().or() // yuck :(
```
elucash: looks entertaining :)

elucash:
👍

criteria/common/src/org/immutables/criteria/constraints/DebugExpressionVisitor.java
asereda-gs: Do you mean something like this ? 

```java
DebugExpressionVisitor(Writer writer) {
   this.writer = new PrintWriter(writer);
}
```

criteria/common/src/org/immutables/criteria/constraints/EmptyExpression.java
asereda-gs: yes it is more a `Nil`

criteria/common/src/org/immutables/criteria/constraints/DebugExpressionVisitor.java
elucash: if I have a `Writer` and debug accepts `PrintWriter` I can always write from the outside, so it's ok the way it is. It can be sometimes useful if we can use more general type inside. I think this is too much overthinking from me for just a debug, sorry ))

criteria/common/test/org/immutables/criteria/ReflectionTest.java
asereda-gs: Yes this is a little ugly. 

Probably in future one can generate two classes (with and without `or()`) something like  
```java
class CriteriaWithoutOr {
}

class Criteria<T>  extends CriteriaWithoutOr<T> {
   public Criteria<T> or() {
   }
}
```

elucash: hm, yep, interesting to think about it

PR #943, opened 2019-04-05
incomplete criteria detection

elucash:
@asereda-gs here's what I was able to came up with wrt criteria detection. It has okeish detection, but in template I've put just to see, and it's not compiling generated code obviously with that `person/Friend` combination. This is not for direct merging obviously, but you can use it as a base

At first, I thought that we can avoid loading protoclass/value type model and get annotation (`CriteriaMirror.isPresent(this.containingTypeElement)` with some additional checks), but then when we need to know package/type name etc, we cannot just guess it, we need style infrastructure etc.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/meta/ValueAttribute.java
+++ b/value-processor/src/org/immutables/value/processor/meta/ValueAttribute.java
@@ -1136,12 +1136,17 @@ public final class ValueAttribute extends TypeIntrospectionBase implements HasSt
     validateThrowsClause();
     validateTypeAndAnnotations();
 
+    initAttributeValueType();
+
     if (supportBuiltinContainerTypes()) {
-      initAttributeValueType();
       initImmutableCopyOf();
     }
 
     initAttributeBuilder();
+  }
+
+  public boolean hasCriteria() {
+    return attributeValueType != null && attributeValueType.isGenerateCriteria();
   }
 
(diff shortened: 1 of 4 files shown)


value-processor/src/org/immutables/value/processor/meta/ValueAttribute.java
```diff
@@ -1136,14 +1136,19 @@ void initAndValidate(@Nullable InstantiationCreator instantiationCreator) {
     validateThrowsClause();
     validateTypeAndAnnotations();
 
+    initAttributeValueType();
```
elucash: putting this into `supportBuiltinContainerTypes()` was probably something short-sighted, but it should work correctly here too

asereda-gs: @elucash thank your for looking into it.
Is it OK if I change `ValueAttribute` ? 
The thing I was missing is :
```java
// inside initAttributeValueType
|| containingType.isGenerateCriteria()
```

asereda-gs:
Thanks, @elucash, I've used this example in #944

value-processor/src/org/immutables/value/processor/meta/ValueAttribute.java
elucash: yep, `ValueAttribute` changes should be ok, the other part is incomplete

PR #948, opened 2019-04-08
Make generated MongoDB repositories extensible

before:
We have use cases in which we want to aggregate documents of generated repositories. Since the generator does not provide an API to build an Aggregation Pipeline, we write them by hand. In this case, we want to extend the generated repository with custom functionality where we need access to the used MongoCollection.

### DIFF
--- a/mongo/src/org/immutables/mongo/repository/Repositories.java
+++ b/mongo/src/org/immutables/mongo/repository/Repositories.java
@@ -99,7 +99,7 @@ public final class Repositories {
       return fieldNamingStrategy;
     }
 
-    private MongoCollection<T> collection() {
+    protected MongoCollection<T> collection() {
       return collection;
     }
 
--- a/mongo/src/org/immutables/mongo/repository/Repositories.java
+++ b/mongo/src/org/immutables/mongo/repository/Repositories.java
@@ -99,7 +99,7 @@ public final class Repositories {
       return fieldNamingStrategy;
     }
 
-    protected MongoCollection<T> collection() {
+    protected final MongoCollection<T> collection() {
       return collection;
     }
 
--- a/value-processor/src/org/immutables/value/processor/Repositories.generator
+++ b/value-processor/src/org/immutables/value/processor/Repositories.generator
@@ -68,7 +68,7 @@ import [starImport];
 [if type allowsClasspathAnnotation 'javax.annotation.concurrent.ThreadSafe']
 @javax.annotation.concurrent.ThreadSafe
 [/if]
-[type.typeDocument.access]final class [type.name]Repository extends Repositories.Repository<[type.typeDocument]> {
+[type.typeDocument.access]class [type.name]Repository extends Repositories.Repository<[type.typeDocument]> {
   private static final String DOCUMENT_COLLECTION_NAME = "[type.documentName]";
 
   private final Serialization serialization;


elucash:
@asereda-gs wdyt, ok?

elucash:
@before thank you for the PR, should be good overall

asereda-gs:
@before can you pls give some examples of how you're planning to extend existing Repository ?

before:
Here is an example that shows simply how we extend the repository with custom aggregations:
```Java
@Immutable
@Repository
abstract class MyValue {
    abstract LocalDate date();
    abstract String text();
}

@Immutable
@TypeAdapters
abstract class SumByDateDto {
    abstract String _id();
    abstract long sum();
}

class MyValueExtendedRepository extends MyValueRepository {

    MyValueExtendedRepository(RepositorySetup configuration) {
        super(configuration);
    }

    AggregateIterable<SumByDateDto> sumByDate(Period period) {
        LocalDate tomorrow = LocalDate.now().plusDays(1);
        LocalDate startInclusive = tomorrow.minus(period);
        LocalDate endExclusive = tomorrow;
        Bson filter = and(
                gt("date", startInclusive),
                lt("date", endExclusive));
        List<Bson> pipeline = Arrays.asList(
                match(filter),
                group("$date", sum("sum", 1)),
                sort(new Document("_id", 1))
        );
        return collection().aggregate(pipeline, SumByDateDto.class);
    }

}
```

We're accessing the protected `collection()` method to execute an aggregation.

@asereda-gs Does it illustrate well enough how we want to use it?

before:
The generated and extended repositories are isolated in a module. I just want to avoid creating custom collection instances if I have already repositories. I don't know what you mean with "not using executor".

asereda-gs:
> I don't know what you mean with "not using executor"

Traditionally repositories returned [FluentFuture](https://google.github.io/guava/releases/23.0/api/docs/com/google/common/util/concurrent/FluentFuture.html). There is `submit(Callable<V>)` method in Repositories for async DB calls. 

@elucash I'm not against this PR but we need to consider such cases in criterias.

elucash:
we would have this accidental consistency of having `protected final Collection` the same way as we have `codecRegistry` or `fieldNamingStrategy` exposed as `protected final`. This could be useful as collection which have coder registry pre-configured by repository. Ideally, I think that just extenally constructing/getting collection and using it directly can be quite "right" solution. However, It can get cumbersome. I think that having protected final collection is quite ok and don't see it as proliferation or  as downside either. In scope of criteria/new repos, agree, that we need to think about how we can provide access to the underlying driver's data structures, but it doesn't look like a pressing issue.

@before can we add `final` to the protected method? will merge then.

before:
@elucash I can live with a `protected final collection()` method. I refined the code to fulfill your request.

asereda-gs:
@before I'll merge it. Can you pls squash the commits so git history is not too cluttered ?

elucash:
@asereda-gs we can squash it during merge, merge pull request button has option in small chevron on the right side.

elucash:
@before thank you!

ftardif:
does this workaround still work with the criteria-api generated repository? I don't seem to have access to getCollection()

asereda-gs:
Since criteria works with different backends (Mongo, Geode, Elasticsearch etc.) there is no direct access to mongo collection. 

That being said you can still create [custom repositories](http://immutables.github.io/criteria.html#custom-repositories) as well as [aggregations](http://immutables.github.io/criteria.html#aggregations).

PR #958, opened 2019-04-13
Make BSON representation of a criteria available

before:
I want to use the BSON representation of a `Criteria` instance in my custom extension of a generated repository.

For example, when I want to build a custom count method in my extended repository I do this:
```Java
@Immutable
@Repository
abstract class MyDoc {
    abstract String name();
}

class MyDocExtendedRepository extends MyDocRepository {

    MyDocExtendedRepository(RepositorySetup configuration) {
        super(configuration);
    }

    long count(String prefix) {
        Criteria criteria = criteria().nameStartsWith(prefix);
        return collection().countDocuments(criteria.toBson());
    }

}
```

### DIFF
--- a/mongo/test/org/immutables/mongo/fixture/criteria/PersonCriteriaTest.java
+++ b/mongo/test/org/immutables/mongo/fixture/criteria/PersonCriteriaTest.java
@@ -248,4 +248,10 @@ public class PersonCriteriaTest {
     check(repository.findAll().fetchAll().getUnchecked()).isEmpty();
     check(repository.find(criteria()).fetchAll().getUnchecked()).isEmpty();
   }
+
+  @Test
+  public void toBson() {
+    check(repository.toBson(repository.criteria())).is(new org.bson.Document());
+    check(repository.toBson(repository.criteria().age(35))).is(new org.bson.Document("age", 35));
+  }
 }
--- a/value-processor/src/org/immutables/value/processor/Repositories.generator
+++ b/value-processor/src/org/immutables/value/processor/Repositories.generator
@@ -43,6 +43,7 @@ import org.immutables.mongo.repository.internal.Support;
 
 import org.bson.codecs.Encoder;
 import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.conversions.Bson;
 [for starImport in type.requiredSourceStarImports]
 import [starImport];
 [/for]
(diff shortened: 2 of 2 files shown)


before:
@elucash and @asereda-gs, any feedback on this if we can get this merged?

elucash:
this seems to be ok, but we might need to decide if we want to expose `toBson` on criteria. Maybe static/or protected method on repository class should be provided (`toBson(criteria)`) @asereda-gs ?

asereda-gs:
@elucash I don't feel strongly either way. 
Another option is to make Criteria implement `Bson` interface (but composition is better than inheritance). 

Do you think long term we should provide `Codec<Criteria>` (at least `Encoder<Criteria>`) ?

elucash:
dunno yet about `Codec<Criteria>`/`Encoder<Criteria>`. For the long term probably. Here I think we can just make `toBson(criteria)` on the repository class, either `static` or `protected` (or both) @asereda-gs

asereda-gs:
@elucash fine with me.
I would initially make it non-static (it can become static later). 

Can we also mark it [@Beta](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/annotations/Beta.html), in case some adjustments are needed ?

before:
I'm confused about the next steps I should do to get it merged. I added the `Beta` annotation to the generated method. Is that enough for now or do you want me to change the code so that the repository class has a `toBson(criteria)` method? If we want to go that route, we can also consider to make the `constraint` field of a `Criteria` instance accessible to implementers via `protected`. Then the caller can use the `Support.convertToBson(constraint)` method for its own. What do you think? Keep as it is now or make the constraint accessible for implementers by providing a protected `constraint()` method?

asereda-gs:
Add `toBson(criteria)` to Repository class. `Support` class is an internal package and shouldn't be leaked outside main Repository API. It is fine to expose `Constraint` instance protected (ideally package private) it is immutable anyway.

before:
I made the suggested changes. Please have a look at it. @elucash @asereda-gs

elucash:
@before looks good, one last question can we just add this method as protected to the abstract repository (that was an idea) and not generating it for every repository?

elucash:
ok I see some non-straightforwardness of this, will augment

elucash:
thanks for PR and updates to it!

PR #959, opened 2019-04-14
Use projection to limit the amount of data that MongoDB sends

before:
To limit the amount of data that MongoDB sends to applications, you can use a projection to specify the fields to include in the resulting documents that match the query filter. A mapper function will be applied to each document in the result set.

```Java
@Mongo.Repository
@Value.Immutable
@Gson.TypeAdapters
interface Person {
    @Mongo.Id
    String id();
    int age();
    List<String> aliases();
    Optional<Date> dateOfBirth();
    Optional<String> middleName();
    String name();
}

class SomeService {
    private final PersonRepository _repository;

    SomeService(RepositorySetup configuration) {
        _repository = new PersonRepository(configuration);
    }

    List<NameAndAge> nameAndAgeListOfPersonsAtLeast30() {
        PersonRepository.Criteria criteria = _repository.criteria().ageAtLeast(30);
        Projection<NameAndAge> projection = Projection.of(ImmutableNameAndAge.class, "name", "age");
        return _repository
                .find(criteria)
                .fetchAll(projection)
                .getUnchecked();
    }
}

@Value.Immutable
interface NameAndAge {
    @Value.Parameter(order = 0)
    String name();
    @Value.Parameter(order = 1)
    long age();
}
```

### DIFF
--- a/mongo/src/org/immutables/mongo/repository/Repositories.java
+++ b/mongo/src/org/immutables/mongo/repository/Repositories.java
@@ -376,8 +376,44 @@ public final class Repositories {
    */
   public static abstract class Projection<T> {
 
+    public static <T> Projection<T> of(final Class<T> resultType, String... fieldsToInclude) {
+      List<String> fields = fieldsToInclude == null
+              ? ImmutableList.<String> of()
+              : ImmutableList.copyOf(fieldsToInclude);
+      return of(resultType, fields);
+    }
+
+    public static <T> Projection<T> of(final Class<T> resultType, List<String> fieldsToInclude) {
+      final Bson fields = fieldsToInclude.isEmpty()
+              ? null
+              : Projections.fields(Projections.excludeId(), Projections.include(fieldsToInclude));
+      return of(resultType, fields);
+    }
+
+    public static <T> Projection<T> of(final Class<T> resultType, @Nullable final Bson projection) {
+      checkNotNull(resultType);
+      return new Projection<T>() {
+        @Nullable
+        @Override
+        protected Gson gson() {
+            return null;
+        }
+
+        @Nullable
+        @Override
+        protected Bson fields() {
+          return projection;
+        }
+
+        @Override
+        protected Class<T> resultType() {
+          return resultType;
+        }
+      };
+    }
+
     /**
-     * @return optional Gson instance to use to be able to to decode to the specified result documents
+     * @return optional {@link Gson} instance to use to be able to to decode to the specified result documents
      */
     @Nullable
     protected abstract Gson gson();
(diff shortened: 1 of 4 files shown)


before:
@elucash and @asereda-gs, any feedback on this if we can get this merged?

elucash:
@before Thank you for PR! I guess it will take time to properly review the design and implementation. Design wise we would be considering the overall approach, then API. Initial design idea was to isolate repository and it's API's from (direct) dependencies on Mongo Java driver APIs (used mostly for configuring and in implementation). Of course we aware that it's not black and white and we already have something accessible or leaked, but still we want to make sure that these APIs would fit original idea as well as the specific need for projections. @asereda-gs please, take a look at this design (probably, not diving into code review) so we can discuss any alternatives or corrections

before:
Important for us is to get more flexibility in writing filters, aggregations and projections. Since the generator is limited I try to find access points to the generated code (and thus saving work). From there we use the MongoDB Java Driver.

asereda-gs:
I'm traveling right now. Will review as soon as I get a chance.

asereda-gs:
@before Thanks for the PR.

Overall, I agree that projections (and aggregations) are needed for mongo repositories.

Regarding current proposal, I think we have to think more about design and consider other use-cases.

Here are some remarks:

### Generate fields (for projection) as opposed to string arguments

Also instead of specifying (projected) fields as strings (see [Projection.of()](https://github.com/immutables/immutables/pull/959/commits/ca943a736fd3b55fb4335bb503861ef573eb95a5#diff-3846314765a239c08f6fe51d301c181cR374)) perhaps generate them as part of repository :

```java
public class MyRepository extends Repositories.Repository<My> {
   public static class Projection extends Repositories.Projection<My> {
        public static final Projection field1;
        public static final Projection field2; 
   }
}
```

Perhaps `Projection` class can be code-gened ? 

### Separate project and fetch functionalities
I think `project()` method should be separated from `fetch` (and friends)  :

```java
// proposed
repository.find(...).project(...).fetchAll();
// instead of 
repository.find(...).fetchAll(projection);
```
There are too many `fetch*` method overloads which makes it confusing for the user.

### Exposing CodecRegistry in Projection

Ideally I would avoid leaking Bson API to the user. Does Projection class need to depend on CodecRegistry ? Why it can't be retrieved from existing collection (via [collection().getCodecRegistry](http://mongodb.github.io/mongo-java-driver/3.9/javadoc/com/mongodb/client/MongoCollection.html#getCodecRegistry--)) ? 

### Complex projections
Should repository support more complex projections like [$project](https://docs.mongodb.com/manual/reference/operator/aggregation/project/) or [$elemMatch](https://docs.mongodb.com/manual/reference/operator/projection/elemMatch/)

These are my initial thoughts (there might be more) and we can discuss them.

before:
Since I have the option extend a generated repository (see #948) there is no need anymore to go the proposed route.

PR #970, opened 2019-05-09
Derived attributes should be transient

### LINKED ISSUE #968: @Derived attributes should be transient
Derived attributes are computed upon construction. Deserialization allows to bypass their computation and potentially allow someone to cheat and construct invalid object by providing specifically crafted input stream.

Unfortunately, it is not enough to mark Derived attributes as transient as they need to be recomputed upon deserialization when constructors are not called.

The solution would be to implement writeReplace/readResolve pair similar to the below code (that I use as a workaround):

```
@MyStyle
@Immutable
@Value.Enclosing
public abstract class VersionInfo implements Serializable
{
    @Immutable
    static abstract class SerialForm
    {

        abstract String versionString();

        final Object readResolve()
        {
            return ImmutableVersionInfo.of(versionString());
        }

    }

    VersionInfo()
    {
    }

    private static final long serialVersionUID = 1L;

    @Auxiliary
    abstract String versionString();

    @Derived
    Version version()
    {
        return Version.parse(versionString());
    }

    final Object writeReplace()
    {
        return ImmutableVersionInfo.SerialForm.of(versionString());
    }

}
```

mkleczek:
Fixes #968 
Work in progress as this is my first time playing with Immutables but I would like to discuss changes/how to contribute etc.

This PR introduces generation of ImmutableSerialForm class and writeReplace() method that returns an instance of this class.
ImmutableSerialForm.readResolve() uses the builder to create the deserialized instance.
readObject() method is also generated in the Immutable value class - it throws StreamCorruptedException to make sure noone tries to cheat by providing an stream with data bypassing the ImmutableSerialForm.

What's left (non-exhaustive list):
- it is not backwards compatible in case the abstract value type defines its own writeReplace or readObject method. I have no idea how to fix this - probably some additional annotation/attribute would be needed.
- it most probably does not handle constructor invocation correctly (just using the builder) and many potential special/corner cases.

Please let me know what to add/change. I am using this implementation in my own project right now and it is suitable **for me** :)

### DIFF
--- a/serial/test/org/immutables/serial/fixture/SerialTest.java
+++ b/serial/test/org/immutables/serial/fixture/SerialTest.java
@@ -42,7 +42,7 @@ public class SerialTest {
 
   @Test
   public void serializablePrehashReadResolve() throws Exception {
-    Method readResolve = ImmutableEnabledWithReadResolve.ImmutableSerialForm.class.getDeclaredMethod("readResolve");
+    Method readResolve = ImmutableEnabledWithReadResolve.class.getDeclaredMethod("readResolve");
     check(readResolve.getModifiers() & Modifier.PRIVATE).is(Modifier.PRIVATE);
   }
 }
--- a/serial/test/org/immutables/serial/fixture/SerialTest.java
+++ b/serial/test/org/immutables/serial/fixture/SerialTest.java
@@ -42,7 +42,7 @@ public class SerialTest {
 
   @Test
   public void serializablePrehashReadResolve() throws Exception {
-    Method readResolve = ImmutableEnabledWithReadResolve.class.getDeclaredMethod("readResolve");
+    Method readResolve = ImmutableEnabledWithReadResolve.ImmutableSerialForm.class.getDeclaredMethod("readResolve");
     check(readResolve.getModifiers() & Modifier.PRIVATE).is(Modifier.PRIVATE);
   }
 }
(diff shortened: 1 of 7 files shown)


mkleczek:
Modified the PR by adding `serialForm` boolean attribute in Immutable annotation that triggers ImmutableSerialForm generation. By default set to false so no code should be affected by this.

elucash:
Thank you for the PR! Will be looking into it. I do see where it's going, but want to dive deeper into concern/leftovers you've mentioned: compatibility, how to opt-in, construction paths (cons/builder) etc

mkleczek:
> 
> 
> Thank you for the PR! Will be looking into it. I do see where it's going, but want to dive deeper into concern/leftovers you've mentioned: compatibility, how to opt-in, construction paths (cons/builder) etc

@elucash - I am happy to work on this if needed so that it could be released ASAP. Right now I'm using patched version of immutables which is less than optimal as you can imagine.

elucash:
Sorry for returning to this after a really long time, I will just leave a note here that I don't think we need to introduce any new mechanisms, seems that similar functionality already implemented in form of `@Serial.Structural` annotation from `org.immutables:serial` module, which will create serialized version with all the `readResolve` `writeReplace` stuff.

mkleczek:
Thanks for your answer. TBH it is a little disappointing. The problem I see with @Serial.Structural annotation is that it does too much. In this case I don't want to amend serialisation format in any way but leave the standard serialisation in place. The only thing is to make sure invariants are met upon deserialization.

PR #985, opened 2019-05-29
Criteria. Migrate matchers to interfaces

asereda-gs:
Matchers are not implemented as interfaces with default methods.


value-processor/src/org/immutables/value/processor/Criteria.generator
```diff
@@ -179,6 +203,19 @@ context.add(Expressions.path("[a.name]"))
...
+[if a.unwrappedElementType eq 'boolean']
+  BooleanMatcher
+[else if a.unwrappedElementType eq 'java.lang.String']
+  StringMatcher
+[else if a.isMaybeComparableKey]
+  ComparableMatcher
+[else]
+  ObjectMatcher
+[/if]
+[/output.trim]
```
elucash: this can give extra newline added, the one after `[/output.trim]` but before `[/template]`

PR #1001, opened 2019-06-07
Add support for optional attributeBuilders.

daicoden:
Currently you get exceptions when using attributeBuilderDetection=true and have an Optional immutable.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -1685,13 +1685,47 @@ checkNotIsSet([v.names.isSet](), "[v.names.raw]");
   [else if v.jdkOptional and ((v.jdkSpecializedOptional or v.optionalAcceptNullable) or v.primitiveElement)]
     this.[v.name] = [maybeCopyOf v][v.name][/maybeCopyOf];
   [else if v.jdkOptional]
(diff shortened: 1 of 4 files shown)


value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -1685,13 +1685,47 @@ checkNotIsSet([v.names.isSet](), "[v.names.raw]");
...
+     * Once called, the attribute builder field is set to a new instance of the builder.
+     * If called more than once, returns the same builder instance.
+     */
+    [eachLine v.builderAttributeAnnotation]
+    [deprecation v]
+    [builderInitAccess v]final [attributeBuilderBuilderType v.getAttributeBuilderDescriptor] [v.names.getBuilder]() {
+      if (this.[v.name] == null) {
+        this.[v.name] = [createAttributeBuilder v.getAttributeBuilderDescriptor];
+      }
+      [nondefaultSetInBuilder v]
```
daicoden: This is directly copied from the non-optional setter flow. I tried to templetize it, but the nondefaultSetInBuilder defined in the let above causes the template to fail. I think you have to pass a function into the template... but can't figure out how to get it to work.

elucash:
Thank you for PR! Actually didn't had time to scrutinize this addition, I use provided tests as main indication that it works and is useful

PR #1004, opened 2019-06-08
Remove generics from Backend interface

asereda-gs:


criteria/geode/src/org/immutables/criteria/geode/GeodeBackend.java
```diff
@@ -23,7 +23,7 @@
 
 import java.util.Objects;
 
-public class GeodeBackend<T> implements Backend<T> {
+public class GeodeBackend<T> implements Backend {
```
elucash: Leftover `<T>` in `GeodeBackend<T>`

PR #1008, opened 2019-06-11
Fix covariance / contravariance in Matchers and Repository

asereda-gs:


value-processor/src/org/immutables/value/processor/CriteriaRepository.generator
```diff
@@ -86,7 +86,7 @@ Repository.Readable<[type.name]>
   /**
    * Find all [type.name](s) given their ids {@code [a.name]}
    */
-  public Repository.Reader<[type.name]> findById(Iterable<? super [a.type]> [a.name]s) {
+  public Repository.Reader<[type.name]> findById(Iterable<? extends [a.type]> [a.name]s) {
```
elucash: Just a random thought: do we need findById(iterable), maybe it can be just with criteria's `where id in (..)` and leave findById only as a convenience for straight findById(ID id) lookup

PR #1011, opened 2019-06-12
Add internal.reactive package with implementations copied from rxjava project

asereda-gs:
This will allow more correct implementations for various reactive operations (like map / flatMap /
flatMapIterable etc.) at the cost of forked code.

The decision between:
1. Simple (home-made) implementation. Possibly with bugs.
2. Copied (but correct) code
3. Hard depedency on rxjava, reactor projects or others. Losing library neutrality.
4. Optional maven dependency
5. [Shade](https://maven.apache.org/plugins/maven-shade-plugin/) rxjava dependency.

is not straight-foward.


asereda-gs:
After discussing with Eugene it was decided to postpone this PR. 
Given number of classes to be copied it is unclear what approach is better.
For now adapters (but not common) will depend directly on rxjava2

PR #1068, opened 2019-08-08
Fix #1067: builtinContainerAttributes doesn't work with JsonAnyGetter

### LINKED ISSUE #1067: builtinContainerAttributes doesn't work with JsonAnyGetter
Following code will fail:

```java
@JsonSerialize
@Value.Style(builtinContainerAttributes = false)
@Value.Immutable
public interface Example {
    @JsonAnyGetter
    @Nullable
    Map<String, JsonNode> getUnknownValues();
}
```

Generated code has wrong arguments for setUnknownValues method:

```java
  @Generated(from = "Example", generator = "Immutables")
  @Deprecated
  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.NONE)
  static final class Json implements Example {
    final Map<String, JsonNode> unknownValues = new HashMap<String, JsonNode>();

    @JsonAnySetter
    public void setUnknownValues(Map<String, JsonNode> key, JsonNode value) {
      this.unknownValues.put(key, value);
    }
    @Override
    public Map<String, JsonNode> getUnknownValues() { throw new UnsupportedOperationException(); }
  }
```

mgorniew:


elucash:
Thank you for PR! Figured out how it works (or rather why it doesn't work before it)

PR #1074, opened 2019-08-21
Enable _id initialization in MongoDB

arouel:
There is no reason to not allow the developer to initialize the `_id` field during upsert.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Repositories.generator
+++ b/value-processor/src/org/immutables/value/processor/Repositories.generator
@@ -393,7 +393,6 @@ public static final class Modifier extends Repositories.Modifier<[type.typeDocum
 [for a in type.allMarshalingAttributes,
     wW = a.wrappedElementType,
     uU = a.unwrappedElementType]
-[if a.marshaledName ne '_id']
   [if a.collectionType]
 
   /**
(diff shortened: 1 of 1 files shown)


before:
@asereda-gs what do you think about that?

asereda-gs:
I think [criteria](https://immutables.github.io/criteria.html) is better :)

asereda-gs:
What are you trying to achieve ?

before:
I want to be able to use `findAndModify()` to create a document with a custom `_id` type and therefore I need to capability to initialize the ID when the document not exists. By removing the condition we enable the generator to provide for all fields initialization methods.

I would give the `criteria` module a chance if there is a release of it. Any plans on this?

asereda-gs:
Yes we have [2.8.0-CriteriaPreview1](https://mvnrepository.com/artifact/org.immutables/criteria-mongo/2.8.0-CriteriaPreview1) already in Maven. `CriteriaPreview2` supporting projections will be released this week. Aggregations are scheduled for next week (as `Preview3`). 

I will check `_id` proposal and get back to you.

elucash:
thank you!  as main focus will be on criteria api from now on, so I'm currently a bit lax on this: if it works then it's ok

PR #1086, opened 2019-09-17
Set optBits for Modifiable, Default maps

### LINKED ISSUE #1055: Modifiable  setter on map does not toggle optBits
The following fails with latest version 2.7.5  : 

```java
  @Test
  public void basic() {
    ModifiableOptBits bits = ModifiableOptBits.create();
    bits.setMap(Collections.emptyMap()); // bit not set
    check(bits.mapIsSet()); // fails

    bits.putAllMap(Collections.emptyMap()); // bit is set
    check(bits.mapIsSet()); // works
  }

  @Test
  public void scary() {
    ModifiableOptBits bits = ModifiableOptBits.create();
    bits.setMap(Collections.singletonMap("one", "1"));
    check(bits.map().keySet()).hasContentInAnyOrder("one"); //  fails
  }

  @Value.Modifiable
  public interface OptBits {
    @Value.Default
    default Map<String, String> map() {
      return Collections.emptyMap();
    }
  }
```
The following is missing on `set*` methods but present on `put*`  : 
```java
    optBits |= OPT_BIT_MAP;
```

Should behavior of `set*()` and `putAll*()` be consistent in terms of `*isSet()` toggle?

dwragge:
Fixes #1055

### DIFF
--- a/value-fixture/src/org/immutables/fixture/modifiable/DefaultMap.java
+++ b/value-fixture/src/org/immutables/fixture/modifiable/DefaultMap.java
@@ -1,0 +1,15 @@
+package org.immutables.fixture.modifiable;
+
+import java.util.Collections;
+import java.util.Map;
+import org.immutables.value.Value;
+
+@Value.Style(jdkOnly = true)
+@Value.Immutable
+@Value.Modifiable
+public interface DefaultMap {
+    @Value.Default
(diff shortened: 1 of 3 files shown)


elucash:
Thank you for PR! I hope this works

Issue #857, opened 2018-10-27
Any call for a cachedHash or lazyHash?


radifalco:
It may already be supported and I just missed it but what I'm wanting is something like a lazy hash. I like prehash but I don't like having to create the hash in every constructor, that's overhead for stuff that will never call `hashCode`. Can we have a `lazyHash` option? Works just like the default but adds a volatile int `hashCode` so that `computeHashCode()` is only called if the volatile field is zero?

elucash:
Thank you for suggestion! It seems that this was requested a while ago, I even checked if this was not already implemented. I cannot easily find those conversations, maybe `prehash=true` was enough for those cases. As about the feature itself, probably, there's no need for `volatile`, it should be fine to recompute hash in presence or race condition assuming that it will be always computed to the same value and that the change to the fields will be observed eventually to all threads. The drawback (probably imaginary) is that unlike the case of prehash there will be no longer the situation where all fields are `final`: I've heard "rumors" of special "immutable" optimizations, but given that this is the way `java.lang.String` is implemented with a lazy hash code, this should not be a big deal. So nowadays I see we can implement this more or less effortless under special `cacheHashCode` style flag. Want to double check the interest in this (?).

Also, for any such piece of functionality, I would kindly ask to provide PR to documentation ;)
(this is a usual recommendation, though from now on I'll be more insisting on this, will have to add a PR template...)

asereda-gs:
I have prepared a PR #1093

PR #1093, opened 2019-09-24
Lazy computation of hashCode #857

asereda-gs:
Add `lazyhash` attribute to `@Value` annotation which generates additional private class field `hashCode`.

```java
// pseudo-code for lazy evaluation of hashcode
public int hashCode() {
  int h = this.hashCode;
  if (h == 0) {
    // 0 is considered unset int
    h = computeHashCode();
    this.hashCode = h;
  }
  return h;
}
```

This pattern is called [racy-single-check](http://javaagile.blogspot.com/2013/05/the-racy-single-check-idiom.html). From Effective Java : 
> If you don't care whether every thread recalculates the value of a field, and the type of the field is a primitive other than long or double, then you may choose to remove the volatile modifier from the field declaration in the single-check idiom [calculating the value of a volatile field if it's null]. This variant is known as the racy single-check idiom. It speeds up field access on some architectures, at the expense of additional initializations (up to one per thread that accesses the field). This is definitely an exotic technique, not for everyday use. It is, however, used by String instances to cache their hash codes.

Also read Jeremy's Mason post [Date-Race-Ful Lazy Initialization for Performance](http://jeremymanson.blogspot.com/2008/12/benign-data-races-in-java.html)

### DIFF
--- a/value-annotations/src/org/immutables/value/Value.java
+++ b/value-annotations/src/org/immutables/value/Value.java
@@ -93,9 +93,18 @@ public @interface Value {
      * In general, use this when {@code hashCode} computation is expensive and will be used a lot.
      * Note that if {@link Style#privateNoargConstructor()} == <code>true</code> this option will be
      * ignored.
+     * For lazy (deferred) {@code hashCode} computation use {@link #lazyhash()}
      * @return if generate hash code precomputing
      */
     boolean prehash() default false;
+
+    /**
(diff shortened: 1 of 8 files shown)


asereda-gs:
How to call the attribute: `lazyhash` vs `lazyHash` (probably later) ?
There is already `prehash`

elucash:
`lazyhash`.  racy-stuff-is-how-I-wanted-to-do-it-too )

PR #1120, opened 2019-11-08
Isuues with Gson Jackson Bridge

bmazzarol:
- Fixed issue where integers are always printed as doubles
- Fixed issue using the jackson XML streaming parser where boolean tokens are always strings

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -71,7 +71,7 @@
       <!-- Optional. Jackson binding have utility used when using polymorhic marshaling with jackson-core implementation -->
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
-      <version>${jackson.version}</version>
+      <version>${jackson-databind.version}</version>
       <scope>provided</scope>
       <optional>true</optional>
     </dependency>
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -28,6 +28,10 @@
     streaming parsing of Gson, while using all of gson binding infrastructure.
   </description>
 
+  <properties>
+    <latest.jackson.databind.version>2.10.0</latest.jackson.databind.version>
+  </properties>
+
   <dependencies>
     <dependency>
       <!-- Required Gson dependency -->
@@ -71,7 +75,7 @@
       <!-- Optional. Jackson binding have utility used when using polymorhic marshaling with jackson-core implementation -->
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
-      <version>${jackson-databind.version}</version>
+      <version>${jackson.version}</version>
       <scope>provided</scope>
       <optional>true</optional>
     </dependency>
@@ -115,6 +119,43 @@
       <groupId>org.glassfish.jersey.containers</groupId>
       <artifactId>jersey-container-grizzly2-http</artifactId>
       <version>2.11</version>
+      <scope>test</scope>
+    </dependency>
+    <!-- Test dependencies to test other jackson serialization formats -->
+    <dependency>
+      <groupId>com.fasterxml.jackson.dataformat</groupId>
+      <artifactId>jackson-dataformat-yaml</artifactId>
+      <version>${jackson.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.dataformat</groupId>
+      <artifactId>jackson-dataformat-xml</artifactId>
+      <version>${jackson.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.woodstox</groupId>
+      <artifactId>woodstox-core</artifactId>
+      <version>5.1.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.dataformat</groupId>
+      <artifactId>jackson-dataformat-smile</artifactId>
+      <version>${jackson.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.dataformat</groupId>
+      <artifactId>jackson-dataformat-properties</artifactId>
+      <version>${jackson.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.dataformat</groupId>
+      <artifactId>jackson-dataformat-cbor</artifactId>
+      <version>${jackson.version}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
@@ -162,6 +203,14 @@
             </configuration>
           </execution>
         </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>8</source>
+          <target>8</target>
+        </configuration>
       </plugin>
     </plugins>
   </build>
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -200,14 +200,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-compiler-plugin</artifactId>
-        <configuration>
-          <source>8</source>
-          <target>8</target>
-        </configuration>
-      </plugin>
     </plugins>
   </build>
 </project>
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -28,10 +28,6 @@
     streaming parsing of Gson, while using all of gson binding infrastructure.
   </description>
 
-  <properties>
-    <latest.jackson.databind.version>2.10.0</latest.jackson.databind.version>
-  </properties>
-
   <dependencies>
     <dependency>
       <!-- Required Gson dependency -->
--- a/gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
+++ b/gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
@@ -142,11 +142,6 @@ public class JsonGeneratorWriter extends JsonWriter implements Callable<JsonGene
     if (value == null) {
       return nullValue();
     }
-    if (!isLenient()) {
-      if (Double.isNaN(value.doubleValue()) || Double.isInfinite(value.doubleValue())) {
-        throw new IllegalArgumentException("JSON forbids NaN and infinities: " + value);
-      }
-    }
     if (value instanceof Integer) {
       generator.writeNumber(value.intValue());
     } else if (value instanceof Short) {
@@ -154,17 +149,33 @@ public class JsonGeneratorWriter extends JsonWriter implements Callable<JsonGene
     } else if (value instanceof Long) {
       generator.writeNumber(value.longValue());
     } else if (value instanceof Float) {
-      generator.writeNumber(value.floatValue());
-    } else if (value instanceof Double) {
-      generator.writeNumber(value.doubleValue());
+      float f = value.floatValue();
+      checkStrictNumber(f);
+      generator.writeNumber(f);
     } else if (value instanceof BigInteger) {
       generator.writeNumber((BigInteger) value);
     } else if (value instanceof BigDecimal) {
-      generator.writeNumber((BigDecimal) value);
+      BigDecimal bd = (BigDecimal) value;
+      checkStrictNumber(bd);
+      generator.writeNumber(bd);
     } else {
-      generator.writeNumber(value.doubleValue());
+      double d = value.doubleValue();
+      checkStrictNumber(d);
+      generator.writeNumber(d);
     }
     return this;
+  }
+
+  private void checkStrictNumber(Number number) {
+    if (!isLenient()) {
+      checkStrictNumber(number.doubleValue());
+    }
+  }
+
+  private void checkStrictNumber(double number) {
+    if (!isLenient() && (Double.isInfinite(number) || Double.isNaN(number))) {
+      throw new IllegalArgumentException("JSON forbids NaN and infinities: " + number);
+    }
   }
 
   @Override
--- a/gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
+++ b/gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
@@ -25,8 +25,6 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.concurrent.Callable;
 
-import static java.lang.Math.floor;
-
 /**
  * {@link JsonWriter} implementation backed by Jackson's {@link JsonGenerator}.
  * Provides measurable JSON writing improvements over Gson's native implementation.
@@ -151,21 +149,19 @@ public class JsonGeneratorWriter extends JsonWriter implements Callable<JsonGene
       }
     }
     if (value instanceof Integer) {
-      generator.writeNumber((Integer) value);
+      generator.writeNumber(value.intValue());
     } else if (value instanceof Short) {
-      generator.writeNumber((Short) value);
+      generator.writeNumber(value.shortValue());
     } else if (value instanceof Long) {
-      generator.writeNumber((Long) value);
+      generator.writeNumber(value.longValue());
     } else if (value instanceof Float) {
-      generator.writeNumber((Float) value);
+      generator.writeNumber(value.floatValue());
     } else if (value instanceof Double) {
-      generator.writeNumber((Double) value);
+      generator.writeNumber(value.doubleValue());
     } else if (value instanceof BigInteger) {
       generator.writeNumber((BigInteger) value);
     } else if (value instanceof BigDecimal) {
       generator.writeNumber((BigDecimal) value);
-    } else if (floor(d) == d) {
-      generator.writeNumber(value.intValue());
     } else {
       generator.writeNumber(d);
     }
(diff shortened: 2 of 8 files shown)


bmazzarol:
Hi Guys,

Thanks for the great library.

I found a few issues using other Jackson data formats (xml, smile, yaml) using the Gson/Jackson bridge.

This resolved the two issues listed above.

Happy to make any changes/add any tests, provide more information, etc...

Cheers
Ben

gson/src/org/immutables/gson/stream/JsonParserReader.java
```diff
@@ -143,7 +144,12 @@ public String nextString() throws IOException {
   @Override
   public boolean nextBoolean() throws IOException {
     requirePeek();
-    boolean value = parser.getBooleanValue();
+    boolean value;
+    try {
+      value = parser.getBooleanValue();
```
asereda-gs: Can you use `parser.getCurrentToken()` to detect JSON type ?

asereda-gs:
Hi Ben,

Thanks for this PR. 

Can you please explain why do you need to write separately  ints and doubles ? What is the usecase ?  

```java
// pseudo-code (simplified logic)
if (isInteger(a)) {
  writer.writeInt(a);
} else {
  writer.writeDouble(a);
}
```

Also is it possible to add tests for this change ?

gson/src/org/immutables/gson/stream/JsonParserReader.java
asereda-gs: Also what is the default behavior of GSON `JsonReader.nextBoolean()`:

```json
{ 
  "foo": true, 
  "bar": "true"
} 
```

will it return `true` in both cases ?

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
```diff
@@ -145,7 +147,12 @@ public JsonWriter value(Number value) throws IOException {
         throw new IllegalArgumentException("JSON forbids NaN and infinities: " + value);
       }
     }
-    generator.writeNumber(d);
+    if (floor(d) == d) {
```
asereda-gs: I think it is better to perform `instanceof`:

```java
if (value instanceof Integer) {
  generator.writeNumber(value.intValue());
} else if (value instanceof Float) {
  generator.writeNumber(value.floatNumber());
} else if (value instanceof Short) {
  generator.writeNumber(value.shortValue());
} else if{
//.... Byte / BigInteger / BigDecimal etc.. 
} else {
  // use double by default
  writer.writeNumber(value.doubleValue());
}
```

gson/src/org/immutables/gson/stream/JsonParserReader.java
bmazzarol: This is not an issue for Gson or the Jackson JsonParser. They both correctly distinguish the token type as boolean. The issue is with the Jackson XML JsonParser. All boolean tokens come come across as a string tokens. So calling `JsonReader.nextBoolean()` Fails here https://github.com/FasterXML/jackson-core/blob/2.11/src/main/java/com/fasterxml/jackson/core/JsonParser.java#L1426-L1431

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
bmazzarol: The reason is because strict parsers that are expecting an integer will fail if a double is found instead. And because it causes a deviation of the output from what Gson produces verses what this produces.

The reason I did not do the instance of checks is because I did not want to have to account for all the possible sub-types of Number. And I did not think the performance would be adversely effected by one call to floor. By my count there is at least 18 classes that directly extend Number in the standard lib, not to mention all the siblings from there.

I will add some unit tests and do a performance comparison between the instance of checks and the comparison with floor.

asereda-gs: I still think it is better to do `instanceof`. For example jackson has special methods for  `BigInteger` / `BigDecimal` for not loosing number precision it is better to leverage them.

Don't forget you're not only doing `Math.floor()` but also `number.doubleValue()` instead of single `instanceof`.

gson/src/org/immutables/gson/stream/JsonParserReader.java
asereda-gs: `JsonReader.nextBoolean()` should only be called when `currentToken` is `VALUE_TRUE` or  `VALUE_FALSE`. 

Perhaps something is wrong elsewhere (like method `getCurrentToken()`) ?

bmazzarol:
@asereda-gs used instanceof checks for all supported Number subtypes, with a catch all that is still integer aware, (uses floor in this case).

I have added unit tests for all the schema-free formats that Jackson has (minus Ion because it is only supported at 2.10). To get serialization working for xml and properties formats I had to add a new reader that expects the producer to only produce string tokens, try using those formats with the JsonParserReader to see the issue.

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
```diff
@@ -145,7 +150,25 @@ public JsonWriter value(Number value) throws IOException {
         throw new IllegalArgumentException("JSON forbids NaN and infinities: " + value);
       }
     }
-    generator.writeNumber(d);
+    if (value instanceof Integer) {
+      generator.writeNumber((Integer) value);
```
asereda-gs: you can use `value.intValue()` here.
The implementation of this method will return `this` (for Integer)

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
```diff
@@ -145,7 +150,25 @@ public JsonWriter value(Number value) throws IOException {
...
+    } else if (value instanceof Short) {
+      generator.writeNumber((Short) value);
```
asereda-gs: same `value.shortValue()` here

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
```diff
@@ -145,7 +150,25 @@ public JsonWriter value(Number value) throws IOException {
...
+      generator.writeNumber((Long) value);
+    } else if (value instanceof Float) {
+      generator.writeNumber((Float) value);
+    } else if (value instanceof Double) {
+      generator.writeNumber((Double) value);
+    } else if (value instanceof BigInteger) {
+      generator.writeNumber((BigInteger) value);
+    } else if (value instanceof BigDecimal) {
+      generator.writeNumber((BigDecimal) value);
+    } else if (floor(d) == d) {
```
asereda-gs: Please move the [statement](https://github.com/immutables/immutables/blob/9d6392258ce5b47b09d7d6b35a5f5b168de3471b/gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java#L142) `double d = value.doubleValue()` (not visible in this review) to where conversion to double is necessary. For instance you don't need conversion to double when you know that current type is `Integer`.

asereda-gs: Let me think more about cases when a double / float `1.0D` needs to be serialized as int `1`. 

One might potentially lose information like positive and negative zero `-0.0F` and `+0.0F`

gson/pom.xml
```diff
@@ -163,6 +200,14 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>8</source>
```
asereda-gs: Can we keep java 7 ? Is there any reason to make it java 8 ? 

I'm not sure if this module is being used by Android developers.

gson/test/org/immutables/gson/bridge/GsonJacksonBridgeSerializationTest.java
```diff
@@ -0,0 +1,247 @@
...
+              try {
+                return factory.createParser(os.toByteArray());
+              } catch (IOException e) {
+                throw new RuntimeException(e);
+              }
+            }
+            , JsonParserReader::new);
+  }
+
+  private <VALUE
```
asereda-gs: This method is hard to follow (5 arguments and 3 generic types). 
Do you need separate `WRITER` / `READER` generic types ? Can one keep them as subclasses to `JsonGenerator` / `JsonParser`

gson/pom.xml
bmazzarol: Sure

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
bmazzarol: Sure

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
bmazzarol: What about the strict validation you were performing?

bmazzarol: More than happy for the else branch to be double. Let me know what you want here

gson/test/org/immutables/gson/bridge/GsonJacksonBridgeSerializationTest.java
bmazzarol: Sure can re-write it.

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
asereda-gs: Yes.  Right now I think double should be last resort (if no other types are matching) :
```java
// many instanceof checks 
} else {
  // last resort serialize as double
 generator.writeNumber(number.doubleValue());
}

```

bmazzarol: Left the `double d = value.doubleValue()` in there to enable isNan strict checks

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
```diff
@@ -145,7 +148,23 @@ public JsonWriter value(Number value) throws IOException {
...
+    } else if (value instanceof Float) {
+      generator.writeNumber(value.floatValue());
+    } else if (value instanceof Double) {
+      generator.writeNumber(value.doubleValue());
+    } else if (value instanceof BigInteger) {
+      generator.writeNumber((BigInteger) value);
+    } else if (value instanceof BigDecimal) {
+      generator.writeNumber((BigDecimal) value);
+    } else {
+      generator.writeNumber(d);
```
asereda-gs: why not `generator.writeNumber(value.doubleValue())` ?

gson/src/org/immutables/gson/stream/JsonParserReader.java
```diff
@@ -143,7 +144,12 @@ public String nextString() throws IOException {
...
+    } catch (JsonParseException e) {
+      value = Boolean.parseBoolean(parser.getValueAsString());
```
asereda-gs: Is your proposed test `GsonJacksonBridgeSerializationTest`  failing without try / catch ?

asereda-gs:
Overall LGTM with some clarifications:
1. Parsing boolean from String (after `JsonParseException`). Want to make sure there are no better alternatives. 
2. Unused `d` variable

gson/src/org/immutables/gson/stream/JsonParserReader.java
bmazzarol: This code can be reverted. It isn't, but is not required as I have subclasses the reader. So to clarify formats that do not support tokens other than string, such as xml and property, will have to use the XmlParserReader.

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
bmazzarol: Fine but in it will mean in cases of a double value with strict checks, doubleValue will Be called twice

asereda-gs: you're talking about `isLinient()`

```java
} else {
   // remove previous "value instanceof Double"
    double d = value.doubleValue();
    if (!isLenient()) {
      if (Double.isNaN(d) || Double.isInfinite(d)) {
        throw new IllegalArgumentException("JSON forbids NaN and infinities: " + value);
      }
    }
   generator.writeNumber(d);
}
```

gson/src/org/immutables/gson/stream/JsonParserReader.java
asereda-gs: Let's keep this logic in `XmlParserReader` then. I would like to avoid making too many assumptions in `JsonParserReader` if that's OK with you.

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
```diff
@@ -139,13 +142,28 @@ public JsonWriter value(Number value) throws IOException {
     if (value == null) {
       return nullValue();
     }
-    double d = value.doubleValue();
     if (!isLenient()) {
```
asereda-gs: @bmazzarol sorry for being pedantic but what I meant is the following:

```java
   if (value instanceof Integer) {
      generator.writeNumber(value.intValue());
    } else if (value instanceof Short || value instanceof Byte) { // write byte as short
      generator.writeNumber(value.shortValue());
    } else if (value instanceof Long) {
      generator.writeNumber(value.longValue());
    } else if (value instanceof Float) {
      generator.writeNumber(value.floatValue());
   // } else if (value instanceof Double) {
   // remove instanceof Double. it is handled in else clause
   //  generator.writeNumber(value.doubleValue());
    } else if (value instanceof BigInteger) {
      generator.writeNumber((BigInteger) value);
    } else if (value instanceof BigDecimal) {
      generator.writeNumber((BigDecimal) value);
    } else {
       // by default, write as double
      // converting to double here (and only here)
     double d = value.doubleValue();
    if (!isLenient() && (Double.isNaN(d) || Double.isInfinite(d))) {
        throw new IllegalArgumentException("JSON forbids NaN and infinities: " + value + " ( for " + value.getClass() + ")");	   
     }	
    generator.writeNumber(d);	    
    }
```

This way you convert to double only once.

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
```diff
@@ -142,9 +142,8 @@ public JsonWriter value(Number value) throws IOException {
...
-      if (Double.isNaN(d) || Double.isInfinite(d)) {
+      if (Double.isNaN(value.doubleValue()) || Double.isInfinite(value.doubleValue())) {
```
elucash: this could be a silly comment on my side, I've noticed that we call `value.doubleValue()` many times, cannot justify the impact because of jit/inlining, just mentioning

asereda-gs: We'll call it only once in `else`.
See my [previous comment](https://github.com/immutables/immutables/pull/1120#discussion_r345316915)

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
bmazzarol: I know what you are asking for, but it is semantically different from the code you had before this pr, and is not correct. If I pass a Double value that is NaN or infinite and strict parsing is enabled no error will be raised.

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
bmazzarol: When strict parsing is enabled, You will have to pay the cost of either a cast, or conversion to double for all Number types that can be NaN or infinite. I'm open to how you want to do that, but can't  speculate on what the JVM does performance wise.

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
asereda-gs: > If I pass a Double value that is NaN or infinite and strict parsing is enabled no error will be raised.

Would it not get into `else` statement and throw exception ? 

Or are you talking about the case when `BigDecimal.doubleValue()` returns `-Infinity` / `+Infinity` ?

gson/src/org/immutables/gson/stream/JsonGeneratorWriter.java
asereda-gs: Why not check only for Floats and Doubles (BigDecimal is a special case) when strict checking is enabled ? 

```java
    System.out.printf("POSITIVE_INFINITY: %s%n", Double.isInfinite(Float.POSITIVE_INFINITY));
    System.out.printf("NEGATIVE_INFINITY: %s%n", Double.isInfinite(Float.NEGATIVE_INFINITY));
    System.out.printf("NaN: %s%n", Double.isNaN(Float.NaN));
```

> POSITIVE_INFINITY: true
> NEGATIVE_INFINITY: true
> NaN: true

asereda-gs: I would say have private method:

```java

public JsonWriter value(Number value) {

   if (value instanceof Float) {
     float f = value.floatValue();
     checkStrictNumber(f);
     generator.writeNumber(f);	
   } else {
      double d = value.doubleValue();
      checkStrictNumber(d);
      generator.writeNumber(d);	
   }
}

void checkStrictNumber(double number) {
  if (isLinient() && (Double.isInfinite(number) || Double.isNaN(number))) {
    throw new IllegalArgumentException("JSON forbids NaN and infinities: " + value);
  }
} 
```

Do you agree with this approach ?

bmazzarol: Yep will implement.

asereda-gs:
Thank you @bmazzarol for your contribution!

Issue #1088, opened 2019-09-19
No arg generic methods fails code generation


yhrn:
If an abstract value class has a generic regular method (i.e. it's not an attribute accossor according to the definition [here](https://immutables.github.io/immutable.html#attributes)) that has a non-void return type and no arguments, Immutables will fail with the message: `Method 'methodName' cannot have own generic type parameters. Attribute accessors can only use enclosing type's type variables`

Below is an example of a (stupid) class that can be used to reproduce the problem on Immutables 2.7.5 (same result using 2.8.0-CriteriaRc0):

```java
package com.company.useless.utils

import java.util.ArrayList;
import java.util.List;
import org.immutables.value.Value;

@Value.Immutable
public abstract class ListFactory {
  abstract int initialCapacity();

  public <T> List<T> create() {
    return new ArrayList<>(initialCapacity());
  }
}
```

elucash:
thank you for reporting! should be a bug

elucash:
Looks like a good fix #1088 . I'll use compilation test from the example provided
```java
import java.util.ArrayList;
import java.util.List;
import org.immutables.value.Value;

@Value.Immutable
public abstract class ListFactory {
  abstract int initialCapacity();

  public <T> List<T> create() {
    return new ArrayList<>(initialCapacity());
  }
}
```

elucash:
fixed in #8ce5911, I'll just see if there's a test

PR #1126, opened 2019-11-21
Attempting to fix problem with generic non attribute methods

yhrn:
See https://github.com/immutables/immutables/issues/1088

I couldn't find any tests for the Value processor so this based just on reading the code and it is completely untested, but I figured it could kick off the discussion.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/meta/AccessorAttributesCollector.java
+++ b/value-processor/src/org/immutables/value/processor/meta/AccessorAttributesCollector.java
@@ -178,9 +178,9 @@ final class AccessorAttributesCollector {
         if (!definingType.equals(originalType) && hasNonInheritedAttributes && nonFinal) {
           report(originalType)
               .warning(About.INCOMPAT,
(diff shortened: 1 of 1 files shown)


asereda-gs:
Take a look at `processor-testlib` module. There is a JUnit [ProcessorRule](https://github.com/immutables/immutables/blob/8c33b68754a4ca2a4fb56fb074ebc552fe5ac201/processor-testlib/src/org/immutables/value/processor/meta/ProcessorRule.java) for `ValueType` / `ValueAttribute` testing. 

Example:  [RepositoryTest](https://github.com/immutables/immutables/blob/8c33b68754a4ca2a4fb56fb074ebc552fe5ac201/criteria/common/test/org/immutables/criteria/processor/RepositoryTest.java)

elucash:
Looks like a good fix #1088 . I'll use compilation test from the example provided
```java
import java.util.ArrayList;
import java.util.List;

@Value.Immutable
public abstract class ListFactory {
  abstract int initialCapacity();

  public <T> List<T> create() {
    return new ArrayList<>(initialCapacity());
  }
}
```

Issue #1006, opened 2019-06-11
Generate constants for properties in Immutables


keesvandieren:
We would like to have properties generated in constants.

Similar to [Lombok FieldNameConstants]( https://projectlombok.org/features/experimental/FieldNameConstants)

cdeszaq:
What is the benefit of this? What does it provide that simply declaring a constant on the abstract type doesn't do?

Additionally, I personally dislike constants that are named the same as their contents, and think they provide no value over using the value directly in code. Constants are best used to provide a name (ie. semantic meaning) to a value, and duplicating the value as the name provides no additional semantic value.

elucash:
I think the overall theme here is having metadata which can be used in various ways and will guard against typos/guessing. If we take a look at just constants for attribute names (putting a full-blown system of value/attribute descriptors aside), the most common need is to use them as values of annotation attributes (of some framework). Annotations attribute only allow the so-called "constant expressions" allowing only putting either literal or inline-able static final constants (initialized with literals). Having that said, there are other quite valid usages for those, as well as some other questionable use-case as you've pointed out.

keesvandieren:
We are using Apache Wicket. There is a [CompoundPropertyModel](https://ci.apache.org/projects/wicket/guide/7.x/single.html#_models_and_javabeans). We've created a customised version of that, that works with Modifiables of immutables framework. Now we have to repeat the propertynames in the Wicket components we make.

stevenschlansker:
Jdbi project also loves to use Immutables but has to guess as to how Builders and spec classes relate.  We'd much rather have proper metadata!

bidorffOL:
I would really love to see this feature integrated in Immutables, it would be really useful in some of our edge cases, but I really prefer the way Lombok handles it to the proposal in #1129: having a `ImmutableX.Fields` do seem better to an separate class. However, in order to have a proper code generation, I think that the generated static fields should respect the code conventions: 

```java
public class ImmutableBean extends Bean {

  private final String myField;
  private final int anotherField;

  [...]

  public static final class Fields {
    public static final String MY_FIELD = "myField";
    public static final String ANOTHER_FIELD = "anotherField";
  }

}
```

PR #1129, opened 2019-12-05
Generate X_Fields.java that simply lists all marshaled fields of X, for every class X marked with @Immutable

DipaanB:
The main driver for this change is to be able to determine the schema of objects (X in our example) that are persisted in a database like mongoDB. Since there is no way to inspect private fields of the Immutable class using reflection, the closest alternative is to inspect the mutator methods on the Immutable Builder class. But that approach is hacky, because the builder class has additional methods (equals, from, wait and collection utility methods like addAll, remove, set, etc.). Also the builder sometimes has overloaded methods, e.g. if a field is of type Optional<T>. Also fields annotated with @Gson.Ignore can't be excluded from the schema, when enumerating mutators on the Builder. Thus generating a new _Fields class seems to be the best available option to be able to determine a database schema.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -20,6 +20,7 @@
 [output.java type.package type.typeImmutable.simple type.element]
 [generateImmutable type true]
 [/output.java]
+[generateFieldsFile type]
       [if type.generateWithInterface]
 [output.java type.package type.typeWith.simple type.element]
 [generateWithInterface type true]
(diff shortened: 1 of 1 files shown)


asereda-gs:
Thank you for this PR.
For similar request pls see #1006 
We're still discussing what is the best way to expose metadata

elucash:
Thank you for PR! but this change kinda falls out of the scope in the way proposed here, however, as mentioned, we're working on alternative. There's experimental `@Data` annotation to expose such metadata. While it's not documented and not stable enough, but it's something that can be tried, will be in the next release (still experimental, maybe eternal beta ;), and also already in version `2.8.0-dt-experimental-7`  here's link to tests to show usage: https://github.com/immutables/immutables/blob/master/data/test/org/immutables/data/DataTest.java

PR #1136, opened 2020-01-09
Fix suppress warnings for generated MongoDB repositories

arouel:
This fixes issues to copy over set custom `@SuppressWarnings` annotations.


asereda-gs:
Thank you!

PR #1149, opened 2020-02-06
instrumented Backend using Micrometer

jmoghisi:
adds an Instrumented Backend using Micrometer Timers.

SELECT Operations are tagged with a parameterized OQL query; there didn't seem to be a better alternative for a human-readable version of the query (SQL, Mongo, etc).


criteria/geode-common/test/org/immutables/criteria/geode/GeodeQueryVisitorTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        check(toOql(person.address.isPresent())).is("address != null");
+        check(toOql(person.address.isAbsent())).is("address = null");
+
+        check(toOql(person.fullName.isEmpty())).is("fullName = $1");
+        check(toOql(person.fullName.notEmpty())).is("fullName != $1");
+    }
+
+    @Test
+    @Disabled("Disabled until collections supported")
```
jmoghisi: I'll add support for Iterable queries, String queries, and Aggregations in a follow up

criteria/micrometer/src/org/immutables/criteria/geode/OqlTagGenerator.java
```diff
@@ -0,0 +1,39 @@
...
+                .pathNaming(ReservedWordNaming.of(PathNaming.defaultNaming()))
+                .useBindVariables(true)
+                .build();
+    }
+
+    @Override
+    public Iterable<Tag> generate(Backend.Operation operation) {
+        if (operation instanceof StandardOperations.Select) {
+            final Query query = ((StandardOperations.Select) operation).query();
+            final Oql oql = oqlGenerator.generate(query);
```
jmoghisi: since the OQL will be parameterised, we should consider adding caching here e.g. using caffeine

asereda-gs:
Thanks a lot for this PR.

As discussed off-line, let's work on Interceptor/Listener API which will be a superset for metrics.
Metrics (micrometer) can then leverage interceptor API to calculate / expose query performance.

PR #1150, opened 2020-02-12
Support OQL Iterable queries

jmoghisi:
Adds support to OQL generator for Iterable queries

### DIFF
--- a/criteria/geode/src/org/immutables/criteria/geode/GeodeQueryVisitor.java
+++ b/criteria/geode/src/org/immutables/criteria/geode/GeodeQueryVisitor.java
@@ -18,6 +18,8 @@ package org.immutables.criteria.geode;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.primitives.Primitives;
+
 import org.immutables.criteria.backend.PathNaming;
 import org.immutables.criteria.expression.AbstractExpressionVisitor;
 import org.immutables.criteria.expression.Call;
@@ -25,6 +27,7 @@ import org.immutables.criteria.expression.ComparableOperators;
 import org.immutables.criteria.expression.Constant;
 import org.immutables.criteria.expression.Expression;
 import org.immutables.criteria.expression.Expressions;
+import org.immutables.criteria.expression.IterableOperators;
 import org.immutables.criteria.expression.Operator;
 import org.immutables.criteria.expression.Operators;
 import org.immutables.criteria.expression.OptionalOperators;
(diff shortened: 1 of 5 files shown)


criteria/common/src/org/immutables/criteria/matcher/OrderMatcher.java
```diff
@@ -27,14 +27,14 @@
   /**
    * Sort current expression in <i>ascending</i> order.
    */
-  default Ordering asc() {
+  default Collation asc() {
```
asereda-gs: Is it possible to keep return type of `asc()` as is (ie `Collation`) ? While it doesn't add functionality it hides all methods from user during IDE auto-complete.

criteria/geode/src/org/immutables/criteria/geode/GeodeQueryVisitor.java
```diff
@@ -165,6 +220,13 @@ private Oql oql(String oql) {
     return new Oql(variables, oql);
   }
 
+  private static String toMethod(Operator op) {
```
asereda-gs: What is the value of this method if it handles just one operator ? 
Maybe remove or add `size` ?

asereda-gs:
Can some `@Disabled` annotations be removed from `GeodeIntegrationTest` ?

criteria/geode/src/org/immutables/criteria/geode/OqlGenerator.java
```diff
@@ -37,23 +34,23 @@
...
   abstract String regionName();
 
-  static OqlGenerator of(String regionName, PathNaming pathNaming) {
+  public static OqlGenerator of(String regionName, PathNaming pathNaming) {
     return ImmutableOqlGenerator.builder().regionName(regionName).pathNaming(pathNaming).build();
   }
 
   @Value.Default
-  boolean useBindVariables() {
+  public boolean useBindVariables() {
```
asereda-gs: Can visibility remain `package-private` ?

criteria/geode/src/org/immutables/criteria/geode/OqlGenerator.java
```diff
@@ -115,6 +113,15 @@ Oql generate(Query query) {
     return new Oql(variables, oql.toString());
   }
 
+  private static boolean isMultiDirectionCollation(Query query) {
+    final long directionCount = query.collations().stream()
+            .map(Collation::direction)
+            .distinct()
+            .limit(Ordering.Direction.values().length)
```
asereda-gs: Will `distinct()` return at most `Ordering.Direction.values().length` anyway ?

criteria/geode/src/org/immutables/criteria/geode/GeodeQueryVisitor.java
```diff
@@ -148,14 +198,19 @@ private Oql binaryOperator(Call call) {
     } else {
       variable = constant.value();
     }
+    return variable;
+  }
 
+  private String getBindVariable(Object variable) {
```
asereda-gs: If this method mutates internal state (see `variables.add()`) I would call it differently (not just getter).

criteria/common/src/org/immutables/criteria/matcher/OrderMatcher.java
jmoghisi: sure, it was needed in test, I will cast for now

criteria/geode/src/org/immutables/criteria/geode/GeodeQueryVisitor.java
jmoghisi: fair point but I will be adding other operators for String (`matches, contains, etc`)

criteria/geode/src/org/immutables/criteria/geode/OqlGenerator.java
jmoghisi: yes, my bad, will address

criteria/geode/src/org/immutables/criteria/geode/OqlGenerator.java
jmoghisi: yes but this short circuits the count to avoid iterating over all collations

criteria/geode/src/org/immutables/criteria/geode/GeodeQueryVisitor.java
jmoghisi: good point, will rename

jmoghisi:
> Can some `@Disabled` annotations be removed from `GeodeIntegrationTest` ?

I don't think any of them cover `Iterable` queries?

asereda-gs:
> Mind pointing me to which integration tests cover `Iterable` queries? I couldn't find any.

Originally I thought that `String.contains`, `String.hasLength` is also part of this PR. If it is not, `GeodeIntegrationTest` can remain as is.

PR #1156, opened 2020-02-14
SuppressedWarnings: if it's deprecated, also suppress forRemoval

stevenschlansker:
Newer Java allows you to mark things as `@Deprecated(forRemoval=true)` which requires a separate warning to be suppressed.

For simplicity's sake, just always add it to the suppress list since it's not harmful if unused.


elucash:
Thank you! I believe it is safe )

Stephan202:
@elucash any chance we can have a new release with this change? :) (I see it's been a while since version 2.8.8 was released :eyes:.)

PR #1162, opened 2020-03-04
Adds support for Fugue Option to CriteriaModel

### LINKED ISSUE #1161: Fugue Option unsupported by Criteria API
After upgrading a project from `Mongo.Repository` to `Criteria.Repository`, the annotation processor fails at this line.

https://github.com/immutables/immutables/blob/7e736f0f478dcefdc2b815bc869a1a766fc5d068/value-processor/src/org/immutables/value/processor/meta/CriteriaModel.java#L206

It appears that Fugue's `Option` is not handled here. Here's a patch which adds support.

clamothe:
Fixes #1161

### DIFF
--- a/criteria/common/pom.xml
+++ b/criteria/common/pom.xml
@@ -91,6 +91,12 @@
       <version>${jackson-databind.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.atlassian.fugue</groupId>
+      <artifactId>fugue</artifactId>
+      <version>3.0.0</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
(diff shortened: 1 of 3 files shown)


criteria/common/pom.xml
```diff
@@ -33,6 +33,11 @@
       <artifactId>guava</artifactId>
       <version>${guava.version}</version>
     </dependency>
+    <dependency>
```
asereda-gs: Is it possible to avoid runtime dependency on fudge ?

asereda-gs:
Thanks for PR. 
Sorry I pushed some changes to `CriteriaModel`. Can you pls rebase with latest master ? 

Although we don't want to support too many optional types this change looks fine because  type can be compared as string (so no direct dependency on atlassian library). 

Btw what backend are you using ?

clamothe:
Thanks. We're using Mongo, albeit haven't yet successfully switched over from `Repository` to `Criteria`.

criteria/common/pom.xml
clamothe: Done

PR #1140, opened 2020-01-19
Add support for Modifiable to extend an inner class/interface

kiddkaffeine:
I was hoping for this feature, and saw several other people had been requesting it, so I decided to take a stab at it.

Turns the current class for the innerBuilder which is extended by Builder and Modifiable implementations.

Supports both classes and interfaces, and will complain if it's a class that does not extend the base Abstract one.

PR #1167, opened 2020-04-02
Unit tests for Modifiable to extend an inner class/interface

kiddkaffeine:
Here are a few simple unit tests illustrating the functionality included in the previous pull request (#1140).

I discover one issue which is fixed here -- all my original test classes had both Value.Immutable and Value.Modifiable annotations;  and this change would fail to kick in if Value.Immutable was set.  This has been corrected, as illustrated in one of the attached tests.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/meta/Constitution.java
+++ b/value-processor/src/org/immutables/value/processor/meta/Constitution.java
@@ -694,6 +694,12 @@ public abstract class Constitution {
       super(names().namings.typeInnerBuilder);
     }
 
+    @Override
+    protected boolean isApplicableTo(Protoclass p) {
+      return p.kind().isValue();
+    }
+
+    @Override
     protected boolean isExtending(TypeElement element) {
       if (element.getKind() == ElementKind.CLASS) {
         String superclassString = SourceExtraction.getSuperclassString(element);
@@ -705,6 +711,7 @@ public abstract class Constitution {
       return false;
     }
 
+    @Override
     protected void lateValidateExtending(TypeElement t) {
       super.lateValidateExtending(t);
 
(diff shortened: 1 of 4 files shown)


value-processor/src/org/immutables/value/processor/meta/Constitution.java
```diff
@@ -694,6 +694,10 @@ public InnerBuilderDefinition() {
       super(names().namings.typeInnerBuilder);
     }
 
+    protected boolean isCorrectKind(Protoclass p) {
```
asereda-gs, line 697: what `correct` in this context means ? 
Can you find a better name and add javadoc.

elucash: I can suggest smth like `isApplicableTo(p)` and briefly document on the abstract method level in the base class. Probably, not much better, but at least smth.

kiddkaffeine: The original base implementation was looking for Value.Immutable annotation, but that's not revenant for Value.Modifiable only classes.  I've adopted @elucash's suggested name for now.

elucash:
👍 thanks for the PR!

kiddkaffeine:
Thanks guys. We're working with the current snapshot now on an update of a large project now and things are looking good.  Out of curiosity, do you have an idea when the next release might be?  (No need to rush, just curious.)

elucash:
no particular plans, we might release it "on demand" rather soon. Probably, I would prefer to roll in some more fixes, but let's say it could happen in 1-2 weeks. thanks!

asereda-gs:
There are a couple of features (for Criteria API) I'd like to add in the next release. As Eugene mentioned, probably a couple of weeks / next month.

PR #1174, opened 2020-05-03
Predictable encoding conflict resolution

### LINKED ISSUE #1170: Inconsistent encoding conflict resolution
I have a type called `Possible` that I'm `@Encoding` to support these types of fields:

- `Possible<T>` 
- `Possible<Optional<T>>`
- `Possible<List<T>>`

However when encountering a class with a combination of these types, I get an incompatibility warning:

```
(immutables:incompat) Encoding conflict for attribute 'field', the winning match:
Possible<T> (from PossibleEncoding).
Other applicable: [Possible<List<T>> (from PossibleListEncoding)]
```

While this warning is expected, the way a winning match is picked changes between builds. I would hope I could assert a certain order of resolution but it doesn't seem to work.

So far my workaround is changing the implementation of `Possible<T>` to `Possible<? extends T>`. Is there another approach to guarantee a given resolution order?

quanticc:
Close #1170

Respect insertion order when iterating multiple encodings by using an ArrayListMultimap to store them.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/encode/Instantiator.java
+++ b/value-processor/src/org/immutables/value/processor/encode/Instantiator.java
@@ -16,7 +16,7 @@
 package org.immutables.value.processor.encode;
 
 import com.google.common.base.Optional;
-import com.google.common.collect.HashMultimap;
+import com.google.common.collect.ArrayListMultimap;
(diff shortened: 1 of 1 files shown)


asereda-gs:
Thank you!

PR #1185, opened 2020-06-03
Truncate strings or values of type variables in `toString` method that are longer than a defined length

arouel:
The idea is to avoid errors when logging immutable objects which carry a lot of data encoded as strings or type variables. This saves us from writing custom `toString` methods to avoid excessive logging when a lot of data gets transported with an immutable to be logged.

Please let me know what you think about it.

### DIFF
--- a/value-annotations/src/org/immutables/value/Value.java
+++ b/value-annotations/src/org/immutables/value/Value.java
@@ -1406,6 +1406,12 @@ public @interface Value {
     Class<? extends Annotation>[] allowedClasspathAnnotations() default {};
 
     /**
+     * Setting to cut strings longer than a defined length when calling the toString method.
+     * @return string limit
+     */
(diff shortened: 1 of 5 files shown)


elucash:
I've played a bit with is, and while making some changes (like avoid generating the util method if not needed etc) there are some test breaks due to error-prone (blows on unused), so I thinking of actually revert to disabling it by default. The main reason is that one of the promises of Immutables was that it's tries to generate quite lean implementation not filled with much of conveniences (and checking for whenever it's needed is not always trivial). I really hope you can use style declaration on some parent package so it will be default for all your types while not changing. I also don't understand that notion of string & type variables, seems like any other objects should be treated the same as type variable substitutions. So I'm changing to handle all objects except for primitives. I hope this works, let me know if not )

Issue #1187, opened 2020-06-11
BoM dependency Management


joke:
Hi,

it would really be great to provide a BoM artifact with a dependency management bundling all the versions of the different artifacts. It would simplify including multiple artifacts.

Something like: [spring-boot-dependencies](https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-dependencies/2.3.0.RELEASE/spring-boot-dependencies-2.3.0.RELEASE.pom) or [jackson-bom](https://repo1.maven.org/maven2/com/fasterxml/jackson/jackson-bom/2.11.0/jackson-bom-2.11.0.pom).

If you are interested I could put something together and make a pull request.

Regards
Joke

elucash:
Can you briefly explain the motivation? We try to avoid using dependencies for the generated code as well as for the generator itself. Most of the dependencies for generated code are optional, the idea is that we try not to drag transitive into user code and don't force specific versions (just hoping that backward compatibility is maintained). Overall we don't mind such contribution, just need to understand for what kind of code it makes sense and that we will not force unnecessary dependencies or being strict with versions to the users of the annotation processor.

joke:
Sure.

Let's say a project is relying heavily on different artifacts:
````
<dependencies>
    <dependency>
        <groupId>org.immutables</groupId>
        <artifactId>value</artifactId>
        <version>2.8.8</version>
        <scope>provided</scope>
    </dependency>
    <dependency>
        <groupId>org.immutables</groupId>
        <artifactId>serial</artifactId>
        <version>2.8.8</version>
        <scope>provided</scope>
    </dependency>
    <dependency>
        <groupId>org.immutables</groupId>
        <artifactId>annotate</artifactId>
        <version>2.8.8</version>
        <scope>provided</scope>
    </dependency>
    <dependency>
        <groupId>org.immutables</groupId>
        <artifactId>mongo</artifactId>
        <version>2.8.8</version>
        <scope>provided</scope>
    </dependency>
    <dependency>
        <groupId>org.immutables</groupId>
        <artifactId>criteria-mongo</artifactId>
        <version>2.8.8</version>
        <scope>provided</scope>
    </dependency>
    <dependency>
        <groupId>org.immutables</groupId>
        <artifactId>criteria-elasticsearch</artifactId>
        <version>2.8.8</version>
        <scope>provided</scope>
    </dependency>
</dependencies>
````
[...]

PR #1191, opened 2020-06-20
Added Bill of Materials (BoM) artifact

joke:
#1187 I had the time to do the required changes needed for a Bill of Materials as described with the issue.

Regards
Joke

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -1,0 +1,176 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Copyright 2014-2018 Immutables Authors and Contributors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ -->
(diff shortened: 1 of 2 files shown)


bom/pom.xml
```diff
@@ -0,0 +1,176 @@
...
+    </parent>
+    <artifactId>bom</artifactId>
+    <name>${project.groupId}.${project.artifactId}</name>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>${project.groupId}</groupId>
+                <artifactId>testing</artifactId>
+                <version>2.8.9-SNAPSHOT</version>
```
Stephan202, line 32: This file will be more maintainable if `2.8.9-SNAPSHOT` is replaced with `${project.version}`.

joke: If you are using the maven release plugin or maven versions plugin maven will automatically set the correct versions.

joke: Should I update the pull request and replace the explicit version?

Stephan202: Let's see what @elucash says, but personally I would, if for no other reason than that (a) it reduces churn and (b) more clearly states intent (namely that these versions should be one and the same.)

We have a company-internal BOM which uses `${project.version}` and have had no issues with that. (The `maven-release-plugin` and `maven-versions-plugin` behave as expected, as do downstream BOM usages.)

joke: As far as I know there are no issues with using `${project.version}` neither with maven nor with gradle. At least not in recent versions.

elucash: My preference would be to use `${project.version}` here: we use vars sparingly, for repeated versions, and I think here is just the right place to use em. Also I stopped using release plugin, use just deploy, a lot less trouble overall and almost the same amount of manual work: replacing versions in editor vs cli. though with vars it will be less occurrences to replace.

joke: Yes replacing it with an editor would be error-prone and tedious. I changed it to `${project.version}`.

elucash: Thank you! that will do.
p.s. In editor, I use search\replace in files, not that difficult and one-off, but having less occurrences and smaller diff helps.

elucash:
Thank you for the PR! merged

bom/pom.xml
joke: Glad I could help.

PR #1208, opened 2020-08-05
Fix SuppressWarnings

arouel:
When adding `@SuppressWarnings("rawtypes")` at method level _Eclipse_ will suppress all raw-types warnings inside that method. The warning `raw` is not supported by any IDE I would know of.


asereda-gs:
Thank you!

PR #1209, opened 2020-08-05
Added toBuilder method to immutables. Closes #1195

### LINKED ISSUE #1195: Why `from` is more sound than `toBuilder`
Could someone please elaborate on the following statement from the documentation? 

`The from method on builders is a much more sound and powerful alternative than having toBuilder() (or alike) on immutable values.`

What are the pros and cons? How is `age.toBuilder().age(42).build()` is really different to `ImmutableAge.withAge(24)`? Both start on immutable value! Second is shorter of course, but that is not the point.

I find it less powerful actually. Because if I have builders and private implementations, I have to do `new AgeBuilder().from(age)`, while `age.toBuilder()` is shorter and could be called by method reference `Age::toBuilder` - you can't do both with `from` approach.

andrey-vasilyev:


value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -4311,6 +4321,27 @@ private static final long serialVersionUID = [literal serialVersion];
...
+      [if v.encoding]
+    .[rr.builderCopyFrom v]([v.names.get]())
+      [else if v.nullable]
+    .[v.names.init]([v.names.get]())
+      [else if v.collectionType]
+    .[v.names.addAll]([v.names.get]())
+      [else if v.mapType]
+    .[v.names.putAll]([v.names.get]())
+      [else]
+    .[v.names.init]([v.names.get]())[/if][/for];
```
andrey-vasilyev, line 4340: There is small issue here. Sometimes semicolon would end up on a separate line (for example, if last attribute is a collection). I am not sure how to deal with it. For iterator has no way to check for last iteration i.e. `for.last` does not exist and can't be easily added (unless enhanced for loop in generators is replaces with iterator-and-while loop). If I missed something (this whole generator/templating framework is quite unusual to me) could you please point in the right direction.

elucash:
thank you for PR, we've implemented `toBuilder` in a slightly different way, via style flag and hard-dependent on regular, non-strict builders, so we always have `from` method used. (available in pre-release `2.8.9-ea-1`) Of course, this PR was a good reminder and example

andrey-vasilyev:
Cool. Thank you!

PR #1211, opened 2020-08-06
Bean friendly modifiable from method return converted result.

SimY4:
Modifiable's `from` method with bean friendly flag turned on returns void in the case where the modifiable class has 1 or more extended interfaces. 

This cause two issuers:

- it's inconsistent with from method that's generated for Modifiable that doesn't extend any extra interfaces
- it causes syntactic errors in Modifiable class in certain scenarios.

This change is focused on fixing these issues.

### DIFF
--- a/value-fixture/src/org/immutables/fixture/modifiable/BeanFriendly.java
+++ b/value-fixture/src/org/immutables/fixture/modifiable/BeanFriendly.java
@@ -27,10 +27,11 @@ import org.immutables.value.Value;
     deepImmutablesDetection = true,
     create = "new",
     get = {"get*", "is*"})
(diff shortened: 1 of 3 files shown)


SimY4:
@elucash Can I also get your opinion on this one? 

This is another inconsistency that I found with bean friendly Modifiables when extra interfaces present.

Without extra interfaces .from method returns self.
With extra interfaces .from method returns void.

That happened because setReturn was used instead for this particular case.

This cause some other bits fail to compile because they expect .from method to return.

elucash:
Good change, thank you! So we're always return this and this makes sense as `from` is unrelated to the setters convention

PR #1212, opened 2020-08-06
When modifiable has modifiable default field, Immutables produces syntactically invalid code.

SimY4:
When modifiable has modifiable default field, Immutables produces syntactically invalid code.

Same is true for derived and lazy.

BEFORE:

```
...
  @Override
  public final ModifiableMod getDefaultMod() {
    return defaultModIsSet()
      ? defaultMod
      : BeanFriendly.super.getDefaultMod(); // doesn't compile because return type of super getDefaultMod is not type of ModifiableMod
  }
...
```

AFTER:

```
...
  @Override
  public final ModifiableMod getDefaultMod() {
    if (defaultModIsSet()) {
      return defaultMod;
    } else {
      BeanFriendly.Mod defaultMod = BeanFriendly.super.getDefaultMod();
      return defaultMod instanceof ModifiableMod
          ? (ModifiableMod) defaultMod
          : new ModifiableMod().from(defaultMod);
    }
  }
...
```

### DIFF
--- a/value-fixture/src/org/immutables/fixture/modifiable/BeanFriendly.java
+++ b/value-fixture/src/org/immutables/fixture/modifiable/BeanFriendly.java
@@ -42,6 +42,39 @@ public interface BeanFriendly {
   @Nullable
   Mod getMod();
 
+  @Value.Default
+  default Mod getDefaultMod() {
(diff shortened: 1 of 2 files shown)


elucash:
thank you for the PR!

PR #1221, opened 2020-09-01
Fix addAll(Iterator<E>) NPE

dmivankov:
For boxed types generated code before this fix contains a loop like
```java
    /**
     * Adds elements to {@link T#getIds() ids} set.
     * @param elements An iterable of ids elements
     * @return {@code this} builder for use in a chained invocation
     */
    @CanIgnoreReturnValue 
    public final T.Builder addAllIds(Iterable<Long> elements) {
      Objects.requireNonNull(elements, "ids element");
      if (this.ids == null) {
        this.ids = new ArrayList<Long>();
      }
      for (long element : elements) {
        this.ids.add(Objects.requireNonNull(element, "ids element"));
      }
      return (T.Builder) this;
    }
```
For invalid calls it is still NPE but static code analysis tools like https://github.com/google/error-prone may rightfully complain that requireNonNull is never needed due to iterating with unboxed type.
--For @SkipNulls case it is an NPE instead of skipping null elements - a more clear bug.--

### DIFF
--- a/value-fixture/src/org/immutables/fixture/nullable/NullablyElements.java
+++ b/value-fixture/src/org/immutables/fixture/nullable/NullablyElements.java
@@ -35,4 +35,10 @@ public interface NullablyElements {
   Map<String, @SkipNulls Integer> sm();
 
   Set<String> rg();
+  
+  Set<@SkipNulls Integer> ri();
+
+  List<@AllowNulls Integer> rj();
+
+  Set<Integer> rk();
 }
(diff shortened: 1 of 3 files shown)


dmivankov:
Will try to add more tests to
https://github.com/immutables/immutables/blob/master/value-fixture/test/org/immutables/fixture/nullable/NullableAttributesTest.java
https://github.com/immutables/immutables/blob/master/value-fixture/src/org/immutables/fixture/nullable/NullablyElements.java

dmivankov:
Actually, skipNull&allowNull cases did work fine, unwrappedElementType was boxed. But without annotations it was unboxed.
So it's only about static code analyzers complaining (blocker to use immutables with bazel 3.4.1 though)

elucash:
Fix makes sense, thank you for PR!

Issue #1223, opened 2020-09-30
Using Immutables.Criteria in Java 9+ environment


realumhelp:
I'm trying to import the Criteria annotations in a JDK11 environment. 

```xml
<dependency>
    <groupId>org.immutables</groupId>
    <artifactId>criteria-inmemory</artifactId>
    <version>2.8.2</version>
</dependency>
```

It seems the module name is not valid because of the dash: `requires org.immutables.criteria-common;`. There are no issues adding the requires statements for Values. Any suggestions would be greatly appreciated!

realumhelp:
So I started looking through some of the code and found the following:

/pom.xml
```xml
              <archive>
                <addMavenDescriptor>false</addMavenDescriptor>
                <manifestEntries>
                  <Automatic-Module-Name>${module.name}</Automatic-Module-Name>
                </manifestEntries>
              </archive>
```
It looks like an automatic module name is created in each jar file. However, if you keep diving down into the criteria modules like **inmemory** or **common**. You'll see that the module.name variable looks to be created from the **project.groupId** and **project.artifactId**. 

/criteria/common/pom.xml
```xml
  <parent>
    <artifactId>criteria</artifactId>
    <groupId>org.immutables</groupId>
    <version>2.8.9-SNAPSHOT</version>
  </parent>
  <modelVersion>4.0.0</modelVersion>
  <artifactId>criteria-common</artifactId>
  <name>${project.groupId}.${project.artifactId}</name>
```
In the above example you'll end up with **module.name = org.immutables.criteria-common**. This is generating an invalid name. 

I'm reaching a bit past my expertise here with Maven and Modules, but this seems to be the issue.

Just for an example sake if I change the /criteria/inmemory/pom.xml's **artifactId** to something like **criteriainmemory** for example it looks like it starts generating a valid module name:

```
jar --describe-module --file ./immutables-master/criteria/inmemory/target/criteriainmemory-2.8.9-SNAPSHOT.jar
No module descriptor found. Derived automatic module.

org.immutables.criteriainmemory@2.8.9-SNAPSHOT automatic
requires java.base mandated
contains org.immutables.criteria.inmemory
```

versus say a delivered JAR from the Maven Repo

```
jar --describe-module --file criteria-inmemory-2.8.8.jar
Unable to derive module descriptor for: criteria-inmemory-2.8.8.jar
Automatic-Module-Name: org.immutables.criteria-inmemory: Invalid module name: 'criteria-inmemory' is not a Java identifier
[...]

PR #1224, opened 2020-10-01
resolve auto-module-name generation

realumhelp:
Fixes immutables/immutables#1223


asereda-gs:
Thanks for PR. I'll take a look at it soon

elucash:
Thank you for PR! With dashes jars are just not working properly anyway, so I guess it's non-controversial fix

PR #1231, opened 2020-10-18
Make Modifiables working with JPA through a new jpaFriendlyModifiable style attribut

### LINKED ISSUE #1230: Make Modifiable more JPA friendly
During my tests with Immutables libs, I encounter some wierd behavious with JPA entities.

It seems that Modifiables is too restrictive to let JPA entities works correctly.

For example with a OneToMany join column 
```
@OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, mappedBy = "opticalStore")
public abstract Set<Employee> getEmployees();
```
Searching data with a `find` (or `findOne` with Spring data) will throw a NPE when populating the `Set<Employee>`. In fact, he should not already load these values but the setter method is generated to erase and populate a LinkedHashSet created early. 
Hopefully, I saw that setting the style `builtinContainerAttributes` to false generate a compliant code but we then lost these useful features. Especially with Hibernate who recommend to manipulate existing collections rather than replace it.

There are another problem with lazy loading. When Searching data with `getReference` (or `getOne` with String data) or when a column is annotated with `@Basic(fetch = FetchType.LAZY)`. Hibernate will override getters method to allow lazy loading. But Modifiables generate final getter methods so this mecanisme isn't possible and entities will only contains null value. 
There are no way to configure Modifiables to generate getter without final modifier.

It would be great to have an option like `beanFriendlyModifiables` to make Modifiable jpa friendly ... `jpaFriendlyModifiables` for example ?

Tiaw:
This is a proposal that fixes #1230 by allowing the `jpaFriendlyModifiables` styles attribut to disable final modifier on getter method and use abstraction for collection rather than the concrete one.

I make an attribut similar to `beanFriendlyModifiables` who does some modification on the generator to make the rendered class compliant for a specific use case. Maybe it can be split in two attributs, one removing the final modifier, the other readjustint collection setters.

### DIFF
--- a/testing/src/org/immutables/matcher/ModifierMatcher.java
+++ b/testing/src/org/immutables/matcher/ModifierMatcher.java
@@ -1,0 +1,36 @@
+package org.immutables.matcher;
+
+import static org.hamcrest.CoreMatchers.not;
+
+import java.lang.reflect.Method;
+
+import org.hamcrest.Matcher;
+import org.immutables.matcher.internal.IsFinal;
+
+/**
+ * Some utilitary methods that check modifier
+ */
+public class ModifierMatcher {
+  /**
+   * Create a matcher for {@link Method} that check it can't be overrided.
+   *
+   * @return the created matcher
+   */
+  public static final Matcher<Object> finalModifier() {
+    return new IsFinal();
+  }
+
+  /**
+   * Create a matcher for {@link Method} that check it can be overrided.
+   *
+   * @return the created matcher
+   */
+  public static final Matcher<Object> notFinalModifier() {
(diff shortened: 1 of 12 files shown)


elucash:
thank you for the PR! It was really hard to decide what to do with it, on one hand jpaFriendly flag would be too specific use case, on the other hand, we have beanFriendlyModifiables for the use case of integrating into frameworks like JPA. Well, the decision we make is to incorporate some parts of this PR into beanFriendlyModifiables. in particular - no final, make field type the same as declaration, not implementation, overwrite collection in setter, etc.

PR #1235, opened 2020-10-20
Write META-INF/services files when annotation processing is over

arouel:
When generating with another annotation processor classes that will be processed by the _Immutables_ generators that also write files in `META-INF/services` the Java Compiler issues a warning like
`javax.annotation.processing.FilerException: Attempt to reopen a file for path /my/project/build/classes/java/main/META-INF/services/com.google.gson.TypeAdapterFactory`. To workaround this problem we should write these files when the annotation processing is over.


elucash:
Thank you! That makes sense. At least I hope it works better for all cases

PR #1244, opened 2020-11-22
replace TravisCI with GitHub Actions CI

sullis:


elucash:
Will it actually work?  Also need to update/remove travisci badge from readme.md

MysterAitch:
GitHub Actions is the way what we've [mostly moved to at JavaParser](https://github.com/javaparser/javaparser/issues/2919), with appveyor/travis still in place.

You should note that this PR is not a 1:1 replacement. This github actions config runs test, which is different to the package / deploy that your travis config previously triggered. I don't know enough about immutable's builds / CI to know if this is important/relevant, though.

Given the config, it should have run on this PR itself and showed up as a status check in the same way that travis did - yet I don't see any checks here on this PR.

(if it were, it would have flagged up the build failure on the latest actions run before merging and in the "checks" tab: https://github.com/immutables/immutables/runs/1439130160#step:5:254 ).

asereda-gs:
It looks like our CI is failing after this merge.

@MysterAitch  can you pls take a look ?

MysterAitch:
> It looks like our CI is failing after this merge.
> 
> @sullis   can you pls take a look ?

#1247

PR #1247, opened 2020-11-23
Immutables to build on all JDKs 8+

SimY4:
I know you didn't ask for it, but now you have it.

I made Immutables build successfully not only on JDK8 but on any other JDK after 8. Your CI shows everything green.

There are some caveats though.

### DIFF
--- a/criteria/common/src/org/immutables/criteria/repository/sync/SyncFetcherDelegate.java
+++ b/criteria/common/src/org/immutables/criteria/repository/sync/SyncFetcherDelegate.java
@@ -53,7 +53,7 @@ class SyncFetcherDelegate<T> implements SyncFetcher<T>, SyncFetcher.DistinctLimi
   @Override
   public Optional<T> oneOrNone() {
     List<T> list = Publishers.blockingListGet(fetcher.oneOrNone());
-    Preconditions.checkState(list.size() <= 1, "Invalid list size: %d", list.size());
+    Preconditions.checkState(list.size() <= 1, "Invalid list size: %s", list.size());
     return list.isEmpty() ? Optional.empty() : Optional.of(list.get(0));
   }
 
--- a/criteria/inmemory/src/org/immutables/criteria/inmemory/ExpressionInterpreter.java
+++ b/criteria/inmemory/src/org/immutables/criteria/inmemory/ExpressionInterpreter.java
@@ -130,7 +130,7 @@ class ExpressionInterpreter implements Function<Object, Object> {
       }
 
       if (op == StringOperators.TO_UPPER_CASE || op == StringOperators.TO_LOWER_CASE) {
-        Preconditions.checkArgument(args.size() == 1, "Size should be 1 but was %d", args.size());
+        Preconditions.checkArgument(args.size() == 1, "Size should be 1 but was %s", args.size());
         Object value = args.get(0).accept(this);
         if (value == null || value == UNKNOWN) {
           return UNKNOWN;
--- a/criteria/pom.xml
+++ b/criteria/pom.xml
@@ -54,18 +54,7 @@
                 <configuration>
                     <compilerArgs>
                         <arg>-XDcompilePolicy=simple</arg>
-                        <arg>
-                            -Xplugin:ErrorProne
-                            -Xep:CheckReturnValue:ERROR
-                            -Xep:MethodCanBeStatic:ERROR
-                            -Xep:BadImport:ERROR
-                            -Xep:MissingOverride:ERROR
-                            -Xep:OrphanedFormatString:ERROR
-                            -Xep:RedundantOverride:ERROR
-                            -Xep:RedundantThrows:ERROR
-                            -Xep:RemoveUnusedImports:ERROR
-                            -Xep:UnusedMethod:ERROR
-                        </arg>
+                        <arg>-Xplugin:ErrorProne -Xep:CheckReturnValue:ERROR -Xep:MethodCanBeStatic:ERROR -Xep:BadImport:ERROR -Xep:MissingOverride:ERROR -Xep:OrphanedFormatString:ERROR -Xep:RedundantOverride:ERROR -Xep:RedundantThrows:ERROR -Xep:RemoveUnusedImports:ERROR -Xep:UnusedMethod:ERROR</arg>
                         <arg>-Xlint:unchecked</arg>
                     </compilerArgs>
                     <annotationProcessorPaths>
--- a/criteria/pom.xml
+++ b/criteria/pom.xml
@@ -52,34 +52,59 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>${maven-compiler-plugin.version}</version>
                 <configuration>
-                    <compilerId>javac-with-errorprone</compilerId>
-                    <forceJavacCompilerUse>true</forceJavacCompilerUse>
                     <compilerArgs>
-                        <arg>-Xep:CheckReturnValue:ERROR</arg>
-                        <arg>-Xep:MethodCanBeStatic:ERROR</arg>
-                        <arg>-Xep:BadImport:ERROR</arg>
-                        <arg>-Xep:MissingOverride:ERROR</arg>
-                        <arg>-Xep:OrphanedFormatString:ERROR</arg>
-                        <arg>-Xep:RedundantOverride:ERROR</arg>
-                        <arg>-Xep:RedundantThrows:ERROR</arg>
-                        <arg>-Xep:RemoveUnusedImports:ERROR</arg>
-                        <arg>-Xep:UnusedMethod:ERROR</arg>
+                        <arg>-XDcompilePolicy=simple</arg>
+                        <arg>
+                            -Xplugin:ErrorProne
+                            -Xep:CheckReturnValue:ERROR
+                            -Xep:MethodCanBeStatic:ERROR
+                            -Xep:BadImport:ERROR
+                            -Xep:MissingOverride:ERROR
+                            -Xep:OrphanedFormatString:ERROR
+                            -Xep:RedundantOverride:ERROR
+                            -Xep:RedundantThrows:ERROR
+                            -Xep:RemoveUnusedImports:ERROR
+                            -Xep:UnusedMethod:ERROR
+                        </arg>
                         <arg>-Xlint:unchecked</arg>
                     </compilerArgs>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>com.google.errorprone</groupId>
+                            <artifactId>error_prone_core</artifactId>
+                            <version>${errorprone.version}</version>
+                        </path>
+                        <path>
+                            <groupId>org.immutables</groupId>
+                            <artifactId>value</artifactId>
+                            <version>${project.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
                 </configuration>
-                <dependencies>
-                    <dependency>
-                        <groupId>org.codehaus.plexus</groupId>
-                        <artifactId>plexus-compiler-javac-errorprone</artifactId>
-                        <version>${errorprone.plexus.version}</version>
-                    </dependency>
-                    <dependency>
-                        <groupId>com.google.errorprone</groupId>
-                        <artifactId>error_prone_core</artifactId>
-                        <version>${errorprone.version}</version>
-                    </dependency>
-                </dependencies>
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>jdk8</id>
+            <activation>
+                <jdk>1.8</jdk>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-compiler-plugin</artifactId>
+                        <configuration>
+                            <fork>true</fork>
+                            <compilerArgs combine.children="append">
+                                <arg>-J-Xbootclasspath/p:${settings.localRepository}/com/google/errorprone/javac/${javac.version}/javac-${javac.version}.jar</arg>
+                            </compilerArgs>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
 </project>
--- a/generator/pom.xml
+++ b/generator/pom.xml
@@ -48,6 +48,14 @@
       <optional>true</optional>
     </dependency>
     <dependency>
+      <!-- only to have compile-time Javac classes -->
+      <groupId>com.google.errorprone</groupId>
+      <artifactId>javac</artifactId>
+      <version>${javac.version}</version>
+      <scope>provided</scope>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
       <groupId>org.immutables</groupId>
       <artifactId>testing</artifactId>
       <version>${project.version}</version>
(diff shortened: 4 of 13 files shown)


criteria/common/src/org/immutables/criteria/repository/sync/SyncFetcherDelegate.java
```diff
@@ -53,7 +53,7 @@ public T one() {
   @Override
   public Optional<T> oneOrNone() {
     List<T> list = Publishers.blockingListGet(fetcher.oneOrNone());
-    Preconditions.checkState(list.size() <= 1, "Invalid list size: %d", list.size());
```
SimY4, line 56: These are actually errorprone errors. It knows that Preconditions doesn't accept anything other than `%s` as a placeholder.

criteria/inmemory/src/org/immutables/criteria/inmemory/ExpressionInterpreter.java
```diff
@@ -130,7 +130,7 @@ public Object visit(Call call) {
       }
 
       if (op == StringOperators.TO_UPPER_CASE || op == StringOperators.TO_LOWER_CASE) {
-        Preconditions.checkArgument(args.size() == 1, "Size should be 1 but was %d", args.size());
```
SimY4, line 133: These are actually errorprone errors. It knows that Preconditions doesn't accept anything other than `%s` as a placeholder.

generator/pom.xml
```diff
@@ -84,5 +76,20 @@
...
+      <activation>
+        <jdk>1.8</jdk>
+      </activation>
+      <dependencies>
+        <dependency>
+          <groupId>com.sun</groupId>
+          <artifactId>tools</artifactId>
+          <version>1.4.2</version>
+          <scope>system</scope>
+          <systemPath>${java.home}/../lib/tools.jar</systemPath>
```
SimY4, line 90: JDK 8 build depends on tools.jar as an artefact.

gson/pom.xml
```diff
@@ -154,6 +154,12 @@
       <version>${jackson.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+      <version>${jaxb-api.version}</version>
```
SimY4, line 160: java xml api was excluded from java distributions starting from Java 9. For jersey to work we need to bring it back.

metainf/pom.xml
```diff
@@ -61,14 +61,6 @@
   </dependencies>
   <build>
     <plugins>
-      <plugin>
-        <artifactId>maven-compiler-plugin</artifactId>
-        <configuration>
-          <compilerVersion>1.6</compilerVersion>
-          <source>1.6</source>
-          <target>1.6</target>
```
SimY4, line 69: Java 11 and higher can't build with Java 6 compat.

value-fixture/src/org/immutables/fixture/custann/CustomEncl.java
```diff
@@ -15,13 +15,17 @@
...
+
 /**
  * The Interface CustomEncl.
  */
 @CustomWithEncStyle
+@Value.Immutable
 public interface CustomEncl {
 
   @CustomWithEncStyle
+  @Value.Immutable
```
SimY4, line 28: I don't know how these were generated before. They have no immutable annotation present anywhere.

value-fixture/src/org/immutables/fixture/custann/CustomImmutables.java
```diff
@@ -15,12 +15,16 @@
...
+import org.immutables.value.Value;
+
 public class CustomImmutables {
 
   @CustomWithStyle
+  @Value.Immutable
   interface Od {}
   
   @CustomWithStyle2
+  @Value.Immutable
```
SimY4, line 27: I don't know how these were generated before. They have no immutable annotation present anywhere.

value-fixture/test/org/immutables/fixture/generatorext/RewriteTest.java
```diff
@@ -15,13 +15,15 @@
...
 
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 public class RewriteTest {
   @SuppressWarnings("CheckReturnValue")
   @Test
+  @Disabled("FIXME: This test doesn't pass on either of provided JDKs")
```
SimY4, line 26: This test says that it should throw IllegalArgumentException but it always throws NullPointerException. On all JDKs. I don't know who to make it pass.

criteria/pom.xml
```diff
@@ -52,34 +52,48 @@
...
-                        <arg>-Xep:MethodCanBeStatic:ERROR</arg>
-                        <arg>-Xep:BadImport:ERROR</arg>
-                        <arg>-Xep:MissingOverride:ERROR</arg>
-                        <arg>-Xep:OrphanedFormatString:ERROR</arg>
-                        <arg>-Xep:RedundantOverride:ERROR</arg>
-                        <arg>-Xep:RedundantThrows:ERROR</arg>
-                        <arg>-Xep:RemoveUnusedImports:ERROR</arg>
-                        <arg>-Xep:UnusedMethod:ERROR</arg>
+                        <arg>-XDcompilePolicy=simple</arg>
+                        <arg>-Xplugin:ErrorProne -Xep:CheckReturnValue:ERROR -Xep:MethodCanBeStatic:ERROR -Xep:BadImport:ERROR -Xep:MissingOverride:ERROR -Xep:OrphanedFormatString:ERROR -Xep:RedundantOverride:ERROR -Xep:RedundantThrows:ERROR -Xep:RemoveUnusedImports:ERROR -Xep:UnusedMethod:ERROR</arg>
```
MysterAitch, line 57: Is there any reason for merging these onto a single line? 

This documentation indicates that newlines inside a single `<arg></arg>` are a problem, but _presumably_ multiple are fine?

> Be aware that when running on JDK 8 the flags cannot be wrapped across multiple lines. JDK 9 and above do allow the flags to be separated by newlines. That is, the second <arg> element above can also be formatted as follows on JDK 9+, but not on JDK 8:
> ```xml
> <arg>
>   -Xplugin:ErrorProne
>   -Xep:DeadException:WARN
>   -Xep:GuardedBy:OFF
> </arg>
> 
> ```

https://errorprone.info/docs/flags

mongo/test/org/immutables/mongo/bson4gson/JavaTimeTest.java
```diff
@@ -93,7 +94,7 @@ public void localDateTime() throws IOException {
 
   @Test
   public void instant() throws IOException {
-    Instant now = Instant.now();
+    Instant now = Instant.now().truncatedTo(ChronoUnit.MILLIS);
```
MysterAitch, line 97: What is the motivation for this change? Is it a java-level specific issue, or more of a tidying/improvement change?

mongo/test/org/immutables/mongo/bson4gson/JavaTimeTest.java
```diff
@@ -75,7 +76,7 @@ public void localDate() throws IOException {
 
   @Test
   public void localDateTime() throws IOException {
-    LocalDateTime now = LocalDateTime.now();
+    LocalDateTime now = LocalDateTime.now().truncatedTo(ChronoUnit.MILLIS);
```
MysterAitch, line 79: What is the motivation for this change? Is it a java-level specific issue, or more of a tidying/improvement change?

pom.xml
```diff
@@ -94,21 +94,22 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <module.name>${project.groupId}.${project.artifactId}</module.name>
     <sonatypeOssDistMgmtSnapshotsUrl>https://oss.sonatype.org/content/repositories/snapshots/</sonatypeOssDistMgmtSnapshotsUrl>
-    <maven.compiler.compilerVersion>1.7</maven.compiler.compilerVersion>
-    <maven.compiler.source>1.7</maven.compiler.source>
-    <maven.compiler.target>1.7</maven.compiler.target>
+    <maven.compiler.compilerVersion>1.8</maven.compiler.compilerVersion>
+    <maven.compiler.source>1.8</maven.compiler.source>
```
MysterAitch, line 98: What is the motivation here? 

If the issue is that Java 11 can't be used to target this version, the github action can be modified to use a lower java version.

elucash:
btw, we maintained java6/7 compatibility for some time mainly for Android. we can solidly go 8+ nowadays it's not needed, people can use older version if really needed etc. it will require releasing next release as 2.9 in my view.
Very much good change, after settling the details we've gladly merge it. Thank you for PR and reviewing!

pom.xml
MysterAitch: @elucash is happy to go with java 8+ support :) 

https://github.com/immutables/immutables/pull/1247#issuecomment-732422495

metainf/pom.xml
MysterAitch: @elucash is happy to go with java 8+ support :) 

https://github.com/immutables/immutables/pull/1247#issuecomment-732422495

criteria/pom.xml
Stephan202: (Author of that documentation here. ;) )

All `-Xplugin:ErrorProne` arguments must be specified inside the same `<arg>`. So on JDK 8 one either has to put them all on one line (like here) or hide the newlines inside XML comments, [like so](https://github.com/PicnicSupermarket/oss-parent/blob/master/pom.xml#L1039-L1079) (you can ignore the `<?SORTPOM` stuff).

More context in the original issue: google/error-prone#1115.

value-fixture/src/org/immutables/fixture/custann/CustomImmutables.java
elucash: It's specifically the test for custom annotation, `@CustomWithStyle` should be the one which replaces both `@Value.Immutable` and the associated style. So something is not working in how resources are loaded/discovered in  java > 8

elucash: Ok for now we can leave with test basically a noop, can you just add FIXME note to either fix or discontinue the feature

value-fixture/test/org/immutables/fixture/generatorext/RewriteTest.java
elucash: the idea here is that generated code have imports rewritten to point to another preconditions 
```java
@Metainf.Service
public class PreconditionsRewriter implements GeneratedImportsModifier {
```

```java
  static <T> T requireNonNull(T instance) {
    if (instance == null) {
      throw new IllegalArgumentException();
    }
    return instance;
  }
```

elucash: seems like `META-INF/services` are not working (any more) is some contexts. (this issue, also custom annotations etc)
Can somebody help a little bit, maybe there are some known information wrt maven/java/annotation processing changes in this area?

criteria/pom.xml
SimY4: On the same page, there's a note that JDK8 doesn't support multiline arguments. Unfortunately in order for this to work on JDK 8 they have to be inlined.

mongo/test/org/immutables/mongo/bson4gson/JavaTimeTest.java
SimY4: Between Java 8 and Java 9 Instant precision changed from millis to nanos. So on JDK9+ we get greater prescision. But when you serialise and deserialize Instants using Gson you still get millis precision.

In order to this test to work on all JDK versions, I eagerly reset precision to millis to unify all JDK builds.

mongo/test/org/immutables/mongo/bson4gson/JavaTimeTest.java
SimY4: Between Java 8 and Java 9 Instant precision changed from millis to nanos. So on JDK9+ we get greater prescision. But when you serialise and deserialize Instants using Gson you still get millis precision.

In order to this test to work on all JDK versions, I eagerly reset precision to millis to unify all JDK builds.

value-fixture/test/org/immutables/fixture/generatorext/RewriteTest.java
elucash: the same here, we can just change fixme to point to `GeneratedImportsModifier` instances are not loaded

elucash: or fix the real issue if we know it

mongo/test/org/immutables/mongo/bson4gson/JavaTimeTest.java
MysterAitch: Today I learned! 

Thank you :)

asereda-gs: Interesting, I thought [Instant](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html) had always nano-second precision:

From javadoc: 
>  the class stores a long representing epoch-seconds and an int representing nanosecond-of-second, which will always be between 0 and 999,999,999.

asereda-gs:
Thank you for this work!
We definitely want to see Immutables build on different JDKs.

Agree that dropping jdk6/7 support requires 2.9 not a patch release

value-fixture/pom.xml
```diff
@@ -172,27 +172,83 @@
...
-          <compilerId>javac-with-errorprone</compilerId>
-          <forceJavacCompilerUse>true</forceJavacCompilerUse>
           <compilerArgs>
-            <arg>-Xep:CheckReturnValue:ERROR</arg>
-            <arg>-Xep:MethodCanBeStatic:ERROR</arg>
+            <arg>-XDcompilePolicy=simple</arg>
+            <arg>-Xplugin:ErrorProne -Xep:CheckReturnValue:ERROR -Xep:MethodCanBeStatic:ERROR -Xep:UnnecessaryCheckNotNull:WARN</arg>
             <arg>-Xlint:unchecked</arg>
           </compilerArgs>
+          <annotationProcessorPaths>
```
elucash, line 180: that is probably why import rewrite and custom annotation extension do not get into classpath of the annotation processing. The question can we switch back to a regular classpath here? Or alternatively we probably need to package test extensions as a separate artifact and then include it here in a annotationProcessorPaths in order for mentioned tests to work.

value-fixture/src/org/immutables/fixture/custann/CustomEncl.java
elucash: probably affected by annotationProcessorPath change

value-fixture/src/org/immutables/fixture/custann/CustomImmutables.java
elucash: probably affected by annotationProcessorPath change

value-fixture/test/org/immutables/fixture/generatorext/RewriteTest.java
elucash: probably affected by annotationProcessorPath change

elucash:
ok, I'll make a module for META-INF stuff in fixture, thanks!

PR #1248, opened 2020-11-24
Support for vavr option.

SimY4:
Javaslang was renamed to vavr quite a while back. Bringing back the support for new names.

### DIFF
--- a/value-fixture/src/org/immutables/fixture/jdkonly/UsingAllOptionals.java
+++ b/value-fixture/src/org/immutables/fixture/jdkonly/UsingAllOptionals.java
@@ -48,6 +48,9 @@ public interface UsingAllOptionals {
   @Value.Parameter
   javaslang.control.Option<String> jso();
 
+  @Value.Parameter
+  io.vavr.control.Option<String> vo();
+
   class Use {
     @SuppressWarnings("CheckReturnValue")
     void use() {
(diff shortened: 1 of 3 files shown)


elucash:
I'm a bit puzzled, have you seen https://github.com/immutables/immutables-vavr , vavr should be addressed by it, or are there specific issues?

elucash:
we can potentially add just this vavr option, just want to make sure that it will not conflict and encoding to take precedence

SimY4:
@elucash Oh! I totally missed this, I'm sorry! Ignore me then, I'll check if immutables-vavr will work for my case. Thank you!

value-fixture/src/org/immutables/fixture/jdkonly/UsingAllOptionals.java
```diff
@@ -59,6 +62,7 @@ void use() {
               .d1(1.1)
               .l1(OptionalLong.empty())
               .jso(javaslang.control.Option.none())
+              .vo(io.vavr.control.Option.none())
```
elucash, line 65: just to note: even if there are precedent with some other variants above, this test is actually testing nothing in that sense that if you remove support for the javaslang.Option you've just added, that accessor will still be generated as this is the default to just have an initializer with the same type as return type. The cases for `v1`, `v2`, and `d1` actually show specially generated initializer like `Builder i1(int int)` for `OptionalInt i1()`, but `Builder i1(OptionalInt i)` is not interesting as it would be generated regardless. So that's the trick

SimY4: It's good to know, thank you

PR #1249, opened 2020-11-25
Fix some previously broken extensions support.

SimY4:
Sorry I broke some extensions. I'm bringing some of them back.

The only problem left is disabled test in https://github.com/immutables/immutables/blob/master/value-fixture/test/org/immutables/fixture/generatorext/RewriteTest.java

### DIFF
--- a/generator/src/org/immutables/generator/AbstractGenerator.java
+++ b/generator/src/org/immutables/generator/AbstractGenerator.java
@@ -80,6 +80,12 @@ public abstract class AbstractGenerator extends AbstractProcessor {
   }
 
   @Override
+  public synchronized void init(ProcessingEnvironment processingEnv) {
+    StaticEnvironment.preInit(processingEnv);
+    super.init(processingEnv);
+  }
+
+  @Override
   public final boolean process(Set<? extends TypeElement> annotations, RoundEnvironment round) {
     try {
       StaticEnvironment.init(annotations, round, processingEnv);
--- a/generator/src/org/immutables/generator/ExtensionLoader.java
+++ b/generator/src/org/immutables/generator/ExtensionLoader.java
@@ -48,16 +48,14 @@ public final class ExtensionLoader {
         List<String> extensions = Lists.newArrayList();
 
         // best effort to read it from compilation classpath
-        if (StaticEnvironment.isInitialized()) {
-          try {
-            String lines = getClasspathResourceText(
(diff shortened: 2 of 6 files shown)


elucash:
I don't know how did you find the fix, but that's marvelous!
Is there is some narrative why the fix is working, I mean that if there anything we can learn about doing or not doing in annotation processors in the future?

elucash:
Thank you! merged

SimY4:
@elucash It was a lot of print lines everywhere in generator code. 😀 

I started from covering [ExtensionLoader](https://github.com/immutables/immutables/pull/1249/files#diff-6e02a9cbc608ce18cccb55da4c9edfa08cadce9d4c2361405105a3c2612d17ca) with print lines because this is where all these META-INF extensions are being picked up. I discovered that pretty much all of the extensions are loading fine except for "org.immutables.value.immutable". This guy has some problems. So I started looking at what is different about it.

What I found is that "org.immutables.value.immutable" is loaded very early in annotation processing lifecycle. And since extension loading is memoized, that means is the first recorded invocation is supposed to be the last one. So I found that that one invocation happens here: https://github.com/immutables/immutables/blob/master/value-processor/src/org/immutables/value/processor/Processor.java#L119

One thing about ExtensionLoader is that it uses ProcessingEnvironment and most of the things and ProcessingEnvironment way is pretty much most effective.

From Processor JavaDoc I learned that the order in which these methods are being invoked is 'init(ProcessingEnvironment)' then 'getSupportedAnnotationTypes' then 'process'. But StaticEnvironment is getting access to ProcessingEnvironment only at 'process' step. Though we can probably provide it earlier at 'init' step. And that's what I did.

asereda-gs:
@SimY4 do you think something like [ProcessorRule](https://github.com/immutables/immutables/blob/master/processor-testlib/src/org/immutables/value/processor/meta/ProcessorRule.java) (or similar idea) can help test/validate future changes to the processor ?

elucash:
thank you for the explanation and it's very interesting! now I don't understand why it worked before (did it?) he he, that is probably only a rhetorical question

SimY4:
@asereda-gs I don't have any experience with these tools to say for sure. I'd love to try but may not find time for this next month.

elucash:
It's ok, I guess or compilation test case revealed the bug, so it did what it should, can leave it as is

asereda-gs:
Fair enough. Thanks for fixing it

PR #1280, opened 2021-03-21
Added eclipse compiler to build matrix

SimY4:
Added eclipse compiler to build matrix. It only affects value-fixture module compilation.

Some fixtures didn't work so I fixed them.
Some fixtures still didn't work and it was much harder to figure out the right fix for them. So i temporarily excluded them from the eclipse compilation.

I also removed so EOLed Java versions from the build matrix to speed things up.


elucash:
thanks!

Issue #1307, opened 2021-05-30
[Encoding] Annotations on custom method parameters are discarded


Alex1304:
Given the following encoding class:

```java
@Encoding
public class CustomOptionEncoding<T> {

    @Encoding.Impl
    private CustomOption<T> value = CustomOption.none();

    @Encoding.Naming("with*OrNull")
    @Encoding.Copy
    public CustomOption<T> withNullable(@Nullable T t) {
        return CustomOption.ofNullable(t);
    }
}
```

Calling `immutableObject.withPropertyOrNull(null)` will raise an IDE warning saying the value cannot be null, because the `@Nullable` annotation is not present in the generated implementation.

Might be related to https://github.com/immutables/immutables/issues/1205

elucash:
thank you for reporting this. We have some limitation with regard to annotation handing in encoding. Transferring annotation from parameter might be possible though, but might require some efforts to properly propagate all the way. All I can promise is to take a look at it within 1–2 weeks. Or is someone could be "brave" enough to help with PR, after all we do copy annotations from encoding methods, so adding param annotation handling might be non-trivial, but not far off the beaten path.

Alex1304:
Just for info, this problem doesn't seem to affect custom Builder methods

PR #1320, opened 2021-07-23
Forward Encoding method parameter annotations

darichey:
Fixes https://github.com/immutables/immutables/issues/1307.

Method parameter annotations in Encodings are now stored in `Param#annotations()`, and the toString/parsing is adjusted to support this with the format `"@Foo @Bar param: Type"`. The Renderers generator is updated to add these annotations to the generated source.

I'm very new to the codebase, so please let me know if you have any suggestions.

### DIFF
--- a/value-processor/src/org/immutables/value/processor/encode/EncodedElement.java
+++ b/value-processor/src/org/immutables/value/processor/encode/EncodedElement.java
@@ -18,6 +18,7 @@ package org.immutables.value.processor.encode;
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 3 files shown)


elucash:
It's a really great work, much appreciated! I did only shallow review but looks really good and thought out. Sorry for delays, being in frequent business trips, attention span rarely allows to dig deeper, but I promise to finish the review and merge in a couple of days

elucash:
Thank you! good stuff

darichey:
@elucash Thanks for merging :) 

Just curious, do you have plans/a timeline for the next release? We're gearing up for a new major release for our library and trying to decide if we should wait for an immutables release with this change.

PR #1326, opened 2021-08-22
Java 16 compat

SimY4:
Making some progress towards Java 17 support.

Right now it only compiles with Java 16. Addresses an issue from https://github.com/immutables/immutables/issues/1283

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,19 +7,21 @@ on:
   push:
     branches:
       - master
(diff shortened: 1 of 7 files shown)


SimY4:
Short summary:
- After you allow illegal access for annotation processing, this project compiles fine.
- value-fixture compiles with no additional setup
- everything mongo related also needs illegal access while test execution. My assumption is that mongo driver itself can't run on Java 16.

PR #1328, opened 2021-08-22
Java 17 compat.

SimY4:
Continuation of https://github.com/immutables/immutables/pull/1326

Looks like everything works on 17-ea also.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ '8', '11', '16' ]
+        java: [ '8', '11', '16', '17-ea' ]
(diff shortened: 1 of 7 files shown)


elucash:
Thank you, will be looking at releasing this

mikebell90:
@SimY4 confused about one thing - you include illegal access for java 16+, but java 17 will ignore that switch.... So immutables CANNOT need it ;)

SimY4:
@mikebell90 it might break if we bump the JDK version to 17 release.

I was just lazy and worked around this problem with a single switch. The long term solution is to add a bunch of --add-opens for all specific JDK modules that the immutable generator requires.

SimY4:
@mikebell90 FYI: https://github.com/immutables/immutables/pull/1337

PR #1340, opened 2021-10-20
Use Java 17 target on Java 17.

SimY4:
- Test Immutables using Java 17 target on Java 17
- Move opens close to build definitions that actually need them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,11 @@ jobs:
       matrix:
         java: [ '8', '11', '17' ]
         compiler: [ 'javac', 'ecj' ]
+        exclude:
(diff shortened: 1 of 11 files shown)


asereda-gs:
Thanks @SimY4 for your PR.

Can you pls give more details in commit comments why these changes are necessary for JDK16+ and not others  ?

asereda-gs:
Also how to maintain `pom.xml` in future when new modules or different dependencies are added ?

elucash:
I understand that something might not be fully simulated & tested, but I'll merge these and will try to fiddle to understand the issue. For us, [not] understanding the impact of the newer releases (16, 17) was one of the reasons to delay the final release of 2.9. So, I'm getting this in and will try to play with the stuff

Issue #1339, opened 2021-10-19
Code generation fails with Java 17


lenaschoenburg:
Over at [Zeebe](github.com/camunda-cloud/zeebe) we are having some issues with `immutables` in combination with `jackson-databind` when upgrading from Java 11 to Java 17. We are using `immutables` version `2.9.0-rc1`.

I'll link a minimal reproducing example below, but the basic issue is this:

Given two classes:

```java
public interface Car {
  List<Tire> getTires();
}

@JsonSerialize
@Immutable
@JsonDeserialize(as = ImmutableCar.class)
public abstract class AbstractCar implements Car {
  @Override
  @JsonDeserialize(as = ImmutableTire.class)
  public abstract List<Tire> getTires();
}
```

and

```java
public interface Tire { }

@Immutable
@JsonDeserialize(as = ImmutableTire.class)
public abstract class AbstractTire implements Tire { }
```

Compilation with Java 11 works as expected but when switching to Java 17 it fails:
<details>
<summary>
Full build log
</summary>
<p>

```
[...]

PR #1341, opened 2021-10-27
fixtures with and without errorprone

SimY4:
This PR aims to reproduce a recently reported issue #1339 .

It can be reproduced with existing set of fixtures if we disable errorprone annotation processor.

It also uncovers a few issues with Java 8 integration that I had to comment out (See the changes in `org.immutables.fixture.nullable.NullableAttributesTest`)


SimY4:
I still don't have a definitive answer on what should happen to avoid these opens. 

My hunch is that this code here no longer flies on Java 17 since jdk.compiler package is not exposed for public use anymore:

https://github.com/immutables/immutables/blob/master/generator/src/org/immutables/generator/Compiler.java

There are a few other places where JDK internal are being used but it all comes down to this detector here.

Issue #1285, opened 2021-04-08
Support for java records


lazystone:
Now with Java 16 we finally  have java records which are native data holders, but what java records lack is the Builder functionality.

It would be nice to add support for Java records, so it's possible to generate Builders for existing java records(as it's done today for interfaces).

elucash:
Thank you for the feature request. I've seen something like this was already opened and then closed as something already working (to some degree). Can you please check if the following solution work:

1. add `org.immutables:builder` artifact dependency (annotations only)
2. Put `@Builder.Constructor` annotation on a record. This annotation would then propagate to record constructor (because of `@Target(ElementType.CONSTRUCTOR)`) and the builder would be, hopefully generated in the same package as it does for Pojos: https://immutables.github.io/factory.html#pojo-constructors

lazystone:
It kind of works actually, but `@Builder.Constructor` has `@Target(ElementType.CONSTRUCTOR)`, so instead of
```java
@Builder
public record MilestoneDumpRecord(
    UUID id, String status, @JsonInclude(Include.NON_ABSENT) Optional<String> text) {}
```

I have to write
```java
public record MilestoneDumpRecord(
    UUID id, String status, @JsonInclude(Include.NON_ABSENT) Optional<String> text) {

  @Builder.Constructor
  public MilestoneDumpRecord(
      UUID id, String status, @JsonInclude(Include.NON_ABSENT) Optional<String> text) {
    this.id = id;
    this.status = status;
    this.text = text;
  }
}
```

Which is kind of nullifies benefits of records. It would be nice just to have some convenience annotation which can be put on the class itself(like `@Builder.Record` maybe?)

lazystone:
However in many cases I probably have to write constructor anyway - to make Optional and List fields non-null...

lazystone:
Yeah... I realized that in order to imitate some of the Immutables behavior for java records I have to write constructors for each record class(default values, checking for nulls), so for me `@Builder.Constructor` works as it is. Closing this.

Thanks for your help @elucash!

cykl:
Something we are loosing with Java records are wither methods. Being able to create an immutable object from scratch using a builder is nice, but if I cannot easily create a new instance derived from an original one it kind of defeat  usefulness of whole thing (i.e. if handwritten derivation is bearable then regular creation is likely fine too).

I just stumbled upon a project that seems to be able to enhance record with wither methods (https://github.com/Randgalt/record-builder#Wither-Example).

[...]

PR #1347, opened 2021-11-24
Emit compile error when using @Value annotations on records

dhoepelman:
As discussed in [#1285](https://github.com/immutables/immutables/issues/1285#issuecomment-930330886) the existing annotations do not work for records. This will remain the case as any support will be introduced as new annotations.

This PR turns the current warning:
```
[WARN] [...] (immutables:incompat) Unmatched annotation will be skipped for annotation processing
```
into an error:
```
[ERROR] [...] error: Value type 'Record1' must be class or interface or annotation type, must be non-final
[ERROR] [...] error: cannot inherit from final Record1
```

Also adds a `tests-java17` module, for testing Immutables with Java 16/17 language features like records and sealed types. Not strictly required for this PR, but necessary for future work on records.

PR sponsored by [Xebia](https://xebia.com/)


tests-java17/test/org/immutables/tests/RecordTests.java
```diff
@@ -0,0 +1,24 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ */
+
+package org.immutables.tests;
+
+public class RecordTests {
+
+    // TODO: Test asserting that @Value.Immutable on records generates a compilation error
```
dhoepelman, line 21: I gave creating a test for this a serious try, but could not find an existing test asserting a compile error and could not get one to work with either the existing ProcessorRule, nor managed to create a new one.

Suggestions welcome

PR #1350, opened 2021-12-09
Fix five flaky tests in value-fixture

pthariensflame:
These were all detected with [NonDex](https://github.com/TestingResearchIllinois/NonDex); these fixes make the tests agnostic to the order of items in collections or entries in JSON objects.

### DIFF
--- a/value-fixture/test/org/immutables/fixture/jackson/BugsTest.java
+++ b/value-fixture/test/org/immutables/fixture/jackson/BugsTest.java
@@ -65,6 +65,7 @@ public class BugsTest {
     check(info.abraCadabra()).is(1);
     check(info.focusPocus());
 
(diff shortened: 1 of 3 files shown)


elucash:
Thank you for the PR! The number of changes is small (not all over the place) and this is good. The reason that we rely on these kind of tests is that the order is predictable in most of cases because of the use of the `ImmutableSet` / `ImmutableMap` from guava which maintain insertion ordering; and also the annotation processor tries hard to maintain consistent ordering in the face of source code declaration ordering and inheritance etc. However, in cases where this is not the case, either `Set.of` `Map.of` collections (or other) are used or even attribute ordering is not consistent under some some circumstances, then it will manifest in flakiness indeed.

PR #1354, opened 2022-01-21
No explicit opens by using Add Opens manifest entry.

SimY4:
Turns out we can include required module patches via manifest entry.


elucash:
Yes, that's a lot better! thank you

Issue #659, opened 2017-07-24
Add 'transient' keyword to hashCode field when prehash is enabled


ghost:
When using  `@Value.Immutable(prehash = true)` `hashCode` is being using by persistent frameworks like spring data.

Adding 'transient' keyword to the transient fields like `hashCode` should solve that I believe.

elucash:
sorry for not responding earlier, the problem with this is that I don't know how or if we should to restore this field upon deserialization. I mean using java binary serialization. Spring data probably only use this modifier to know if it should skip the field. But in order to do consistent functionality we need to check if deserialization paths can recreate consistent objects: we have some custom serialization proxies generated which potentially restore all the correct state, but I'm not sure 100%

andrey-cabonline:
If I understand problem correctly, you should bother about serialization/deserialization only if class implements java.io.Serializable.
If it does not then it's safe to apply transient keyword.

If it does implement java.io.Serializable then generated class might have something like:
```java
private void readObject(java.io.ObjectInputStream in) throws IOException, ClassNotFoundException {
    in.defaultReadObject();
    setEageryCalculatedFields();
}
```

More on [readObject](https://docs.oracle.com/javase/8/docs/api/java/io/Serializable.html).

PR #1369, opened 2022-03-02
Using (lazyhash = true) should always make the computed hashCode field transient

### LINKED ISSUE #1368: Using `(lazyhash = true)` with `Serializable` makes computed `hashCode` field non-transient
When using `(lazyhash = true)` one would expect that the generated `hashCode` field to store the computed hashcode would be `transient`.

This is actually the case for the example above with `LazyHash`:

```
@Value.Immutable(lazyhash = true)
public interface LazyHash {
  String s();
  boolean b();
  int i();
}
```

One can see that `hashCode` is indeed `transient`:

```
public final class ImmutableLazyHash implements LazyHash {
  private final String s;
  private final boolean b;
  private final int i;
  @SuppressWarnings("Immutable")
  private transient int hashCode; // hashCode lazily computed
  ....
  ....

}
```

However, I would have expected the same behavior when making a class/interface `Serializable`:

```
@Value.Immutable(lazyhash = true)
public interface LazyHashSerializable extends Serializable {
  String s();
  boolean b();
  int i();
}
```

But the generated class actually shows that `hashCode` is not `transient`:

```
public final class ImmutableLazyHashSerializable implements LazyHashSerializable {
  private final String s;
  private final boolean b;
  private final int i;
  @SuppressWarnings("Immutable")
  private int hashCode; // hashCode lazily computed
  ....
  ....
}
```

I'll try to provide a PR to address this.

nastra:
Fixes #1368

When using `(lazyhash = true)` one would expect that the generated `hashCode` field to store the computed hashcode would be `transient`.

This is actually the case for the example above with `LazyHash`:

```
@Value.Immutable(lazyhash = true)
public interface LazyHash {
  String s();
  boolean b();
  int i();
}
```

One can see that `hashCode` is indeed `transient`:

```
public final class ImmutableLazyHash implements LazyHash {
  private final String s;
  private final boolean b;
  private final int i;
  @SuppressWarnings("Immutable")
  private transient int hashCode; // hashCode lazily computed
  ....
  ....

}
```

However, I would have expected the same behavior when making a class/interface `Serializable`:

```
@Value.Immutable(lazyhash = true)
public interface LazyHashSerializable extends Serializable {
  String s();
  boolean b();
  int i();
}
```

But the generated class actually shows that `hashCode` is not `transient`:

```
public final class ImmutableLazyHashSerializable implements LazyHashSerializable {
  private final String s;
  private final boolean b;
  private final int i;
  @SuppressWarnings("Immutable")
  private int hashCode; // hashCode lazily computed
  ....
  ....
}
```


nastra:
@elucash could you review this please?

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -2512,7 +2512,7 @@ this.[n] = [valueFromValue v][invokeSuper v].[v.names.get]()[/valueFromValue];
   [eachLine type.syntheticFieldsInjectedAnnotations]
   [jsonIgnore type]
   [if type.useLazyhash][hiddenMutableState type][/if]
-  private[if not type.serial.simple] transient[/if][if type.usePrehashed] final[/if] int hashCode;[if type.useLazyhash] // hashCode lazily computed[/if]
+  private[if ((not type.serial.simple) or type.useLazyhash)] transient[/if][if type.usePrehashed] final[/if] int hashCode;[if type.useLazyhash] // hashCode lazily computed[/if]
```
nastra, line 2515: I was thinking to remove the entire if check around `transient` but I guess we can't do that judging by the conversation in https://github.com/immutables/immutables/issues/659 @elucash?

elucash: I was struggling to remember why `not type.serial.simple` check was put in the first place. Probably it is for precomputed hash code to be deserialized correctly when using simple(regular) serialization. So it should be correct to force transient for the `useLazyhash`. So far it seems correct solution.

nastra:
thanks for the quick review @elucash :)

PR #1371, opened 2022-03-07
Improve error msg for null values in JDK Maps

nastra:
This slightly improves the error message for null values in JDK maps to
also mention the key, as this makes it easier to figure out for which
key a null value was passed


nastra:
@elucash could you review this one please?

value-fixture/test-java-11/org/immutables/fixture/nullable/NullableAttributesTest.java
```diff
@@ -330,4 +333,32 @@ public String b() {
...
     check(s.b()).is("b");
   }
+
+  @Test
+  public void nullValuesInJdkMapErrorMsg() {
+    Map<String, String> properties = new HashMap<>();
+    properties.put("a", "a val");
+    properties.put("b", null);
+    properties.put("c", "c val");
+    try {
```
nastra, line 343: @elucash would it make sense to introduce [AssertJ](https://assertj.github.io/doc/) to the project to have more fluent assertion checks? This particular example could then be simplified to 
```
assertThatThrownBy(() -> ImmutableEntityWithMap.builder().properties(properties).build())
        .isInstanceOf(NullPointerException.class)
        .hasMessage("properties value for key: b");
```
There are also a lot of other good examples around Exception checks in https://github.com/assertj/assertj-examples/blob/main/assertions-examples/src/test/java/org/assertj/examples/ExceptionAssertionsExamples.java.

Let me know what you think. If you think this could be useful, then I could provide a PR with some example test code to show AssertJ usage and how it could make writing tests more fluent/easier to check.

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -4239,7 +4239,7 @@ private static <K, V> java.util.Map<K, V> createUnmodifiableMap(boolean checkNul
     V v = e.getValue();
     if (checkNulls) {
       [requireNonNull type](k, "key");
-      [requireNonNull type](v, "value");
+      [requireNonNull type](v, "value for key: " + k);
```
asereda-gs, line 4242: I wonder if string concatenation can be costly here ?
Perhaps use a version with the [Supplier](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.util.function.Supplier-) ?

value-fixture/test-java-11/org/immutables/fixture/nullable/NullableAttributesTest.java
asereda-gs: We have discussed using AssertJ or Google's Truth assertions frameworks in immutables before. For now we have decided to still use `check()` API.

value-processor/src/org/immutables/value/processor/Immutables.generator
nastra: I tried using `[requireNonNull type](v, () -> "value for key: " + k);` already but that change is slightly more involved as it probably requires to change a bunch of other places. So maybe this can be done in a follow-up?

elucash: I was thinking the same, I cannot predict exact impact, but most of the precondition checks are highly optimized for almost zero cost noop case, this up-front concatenation have high chance to skew inlining etc. Maybe we can just prefix it with, `if (v == null) `, like

```java
if (v == null) [requireNonNull type](v, "value for key: " + k)
```

elucash: we will pay for branching in this case, guess it's an acceptable tradeoff

nastra: fixed

value-processor/src/org/immutables/value/processor/Immutables.generator
```diff
@@ -1872,7 +1872,7 @@ checkNotIsSet([v.names.isSet](), "[v.names.raw]");
       [else]
     this.[v.name].put(
         [requireNonNull type](key, "[v.name] key"),
-        [requireNonNull type](value, "[v.name] value"));
+        [requireNonNull type](value, "[v.name] value for key: " + key));
```
nastra, line 1875: @elucash how can I check at this place whether `value` is a primitive or not? I was experimenting with 
```
[if not v.primitiveElement]value == null ? [/if][requireNonNull type](value, "[v.name] value for key: " + key)[if not v.primitiveElement] : value[/if]);
```

but it would produce not the results I was hoping for as it failed later here (since you can't do a null check without boxing):

```
public final JdkMapsBuilder putJust(long key, int value) {
    this.just.put(
        Objects.requireNonNull(key, "just key"),
        value == null ? Objects.requireNonNull(value, "just value for key: " + key) : value);
    return this;
}
```

elucash, line 1875: this one should be changed too, or was this one left out intentionally? thank you

nastra: see my question above about how to check whether value is a primitive or not

elucash: yes, reading and thinking, thank you

elucash: `v.primitiveElement` is only for keys, so it ignores values. I've found some duplicated methods(accessors) and among those there seems to be these ones which are more appropriate

* `v.unwrappedElementPrimitiveType` - if key would be primitive in `put(key, value)`, works on 1st type parameter of a map
* `v.unwrappedSecondaryElementPrimitiveType` - if value would be primitive in `put(key, value)`, works on 2nd type parameter of a map

So in this, you would just use `unwrappedSecondaryElementPrimitiveType` check

nastra: thanks for your help, that worked.


I adjusted the condition to 
```
[if v.unwrappedSecondaryElementPrimitiveType]value[else]value == null ? [requireNonNull type](value, "[v.name] value for key: " + key) : value[/if])
```
so we're not doing a null check on a primitive and hence avoiding the cost of boxing. Let me know what you think.

elucash:
Good job, thank you for PR. CI just completed, merging it

PR #1406, opened 2022-09-29
Avoid invalidating Gson type adapter cache when creating Mongo Repository

jmoghisi:
Avoid invalidating Gson type adapter cache when creating Mongo Repository.

We only register the delegatingTypeAdapterFactory if Gson does not have TypeAdapters registered for all Bson related types

### DIFF
--- a/mongo/src/org/immutables/mongo/bson4gson/GsonCodecs.java
+++ b/mongo/src/org/immutables/mongo/bson4gson/GsonCodecs.java
@@ -19,9 +19,11 @@ import com.google.common.base.Preconditions;
 import com.google.gson.Gson;
 import com.google.gson.TypeAdapter;
 import com.google.gson.TypeAdapterFactory;
+import com.google.gson.internal.bind.ReflectiveTypeAdapterFactory;
 import com.google.gson.reflect.TypeToken;
 import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonWriter;
+import com.mongodb.MongoClientSettings;
 import org.bson.AbstractBsonReader;
 import org.bson.codecs.Codec;
 import org.bson.codecs.DecoderContext;
@@ -150,9 +152,35 @@ public final class GsonCodecs {
     };
   }
 
+  /**
+   * Appends a TypeAdapterFactory to allow Gson and Bson adapters to co-exist.
+   * Avoids registering the TypeAdapterFactory if Gson already has factories for BSON types.
+   *
+   * @param gson preconfigured instance
+   * @return {@code gson} configured with delegating type adapter factory.
+   */
+  public static Gson newGsonWithBsonSupport(Gson gson) {
+    // Will be used as a factory for BSON types (if Gson does not have one). By default, uses
+    // TypeAdapter(s) from Gson if they're explicitly defined (not a ReflectiveTypeAdapter).
+    // Otherwise delegate to BSON codec.
+    TypeAdapterFactory bsonAdapterFactory = delegatingTypeAdapterFactory(
+        MongoClientSettings.getDefaultCodecRegistry()
+    );
+
+    // Appending new TypeAdapterFactory to allow Gson and Bson adapters to co-exists.
+    // Depending on the type we may need to use one or another. For instance,
+    // Date should be serialized by Gson (even if Bson has codec for it).
+    // But ObjectId / Decimal128 by BSON (if Gson doesn't have a type adapter for it).
+    // Document or BsonDocument should only be handled by BSON (it's unlikely that users have direct dependency on them in POJOs).
+    // So newGson is a way to extend existing Gson instance with "BSON TypeAdapter(s)"
+    return gson.newBuilder()
+        .registerTypeAdapterFactory(bsonAdapterFactory)
+        .create();
+  }
+
   static <A> boolean isReflectiveTypeAdapter(TypeAdapter<A> adapter) {
     Preconditions.checkNotNull(adapter, "adapter");
-    return adapter instanceof com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.Adapter;
+    return adapter instanceof ReflectiveTypeAdapterFactory.Adapter;
   }
 
   /**
--- a/mongo/src/org/immutables/mongo/repository/RepositorySetup.java
+++ b/mongo/src/org/immutables/mongo/repository/RepositorySetup.java
@@ -24,7 +24,6 @@ import com.google.gson.GsonBuilder;
 import com.google.gson.TypeAdapterFactory;
 import com.mongodb.DB;
 import com.mongodb.MongoClient;
-import com.mongodb.MongoClientSettings;
 import com.mongodb.MongoClientURI;
 import com.mongodb.ReadPreference;
 import com.mongodb.WriteConcern;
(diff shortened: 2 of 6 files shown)


mongo/src/org/immutables/mongo/bson4gson/GsonCodecs.java
```diff
@@ -150,9 +169,43 @@ public <T> Codec<T> get(Class<T> clazz) {
...
 
+  /**
+   * Appends a TypeAdapterFactory to allow Gson and Bson adapters to co-exist.
+   * Avoids registering the TypeAdapterFactory if Gson already has factories for BSON types.
+   *
+   * @param gson preconfigured instance
+   * @return {@code gson} configured with delegating type adapter factory.
+   */
+  public static Gson newGsonWithDelegatingTypeAdapterFactory(Gson gson) {
+    if (hasBsonTypeAdapters(gson)) {
```
jmoghisi, line 180: @asereda-gs not too happy with this approach but could not identify a better way to detect if these type adapters were already registered. an alternative I considered was to add another builder method to `RepositorySetup` that allows providing a preconfigured Gson instance. please let me know your thoughts.

elucash:
recovering from Ian storm, will get back in a couple of days, sorry

talios:
On 4/10/2022 at 9:05:58 PM, Eugene Lukash ***@***.***> wrote:

> recovering from Ian storm, will get back in a couple of days, sorry
>
>
Off topic - but glad to hear your ok, and hope you weren’t too
displaced/affected - looks beyond nasty!

Mark

-- 
"Great artists are extremely selfish and arrogant things" — Steven Wilson,
Porcupine Tree

elucash:
Thank you Mark @talios for support! We were lucky! No comments
IMG_2982
IMG_2988
IMG_2981

elucash:
only minor damage, happy to survive this shit ))

elucash:
@jmoghisi I've reviewed the PR, and I think I've got the idea. The `mongo` module is in maintenance mode mostly, newer stuff went into `criteria` module, including MongoDB backend implementation. I'm ok with these changes overall, but want to minimize the surface of changes. I would propose to remove the changes (and cleanup other corresponding methods) in `RepositorySetup.Builder.gson(` and instead create a new method `RepositorySetup.Builder.preconfiguredGson(` where we will  take `Gson` instance as is, essentially what `if (hasBsonTypeAdapters(gson)) {...` branch is doing. In light of this, it's even possible to get away without this new method, because we already can pass `Builder codecRegistry(CodecRegistry registry)` which is just 2 calls away from `preconfiguredGson`  via `CodecRegistries.fromRegistries(GsonCodecs.codecRegistryFromGson(`.

So in the dry down we could have just some new codecs registered, some changes to util, constants, and tests.  Plus, if you really wish – add `RepositorySetup.Builder.preconfiguredGson(` or leave it out if favor of calling `RepositorySetup.Builder.codecRegistry` and construct gson/codecs manually.

Minor stuff - if possible change line continuation indent to 4 spaces instead of 8 (even if we had it accidental 8 previously). The regular indent is 2 which is correct. Also make `);` trail on the previous line, without line break

```java
  smth = wrappingCall(new Something.Something()
      .lineContinuation()  // <-- +4 continuation indent
      .anotherLine());  // <-- no newline before closing );
```

asereda-gs:
hi @jmoghisi, thanks for this PR.

I tend to agree with Eugene. Maybe modify RepositorySetup to provide "raw" Gson or have a boolean flag to indicate not to enhance existing gson instance.

Also we you can't leverage `Builder.codecRegistry()` for your usecase ?

jmoghisi:
Thanks you both for the feedback. I've removed the changes to `Builder#gson`.

`Builder#codecRegistry` should be sufficient but I just needed to made `GsonNamingStrategy` public.

mongo/src/org/immutables/mongo/bson4gson/GsonCodecs.java
```diff
@@ -150,9 +152,35 @@ public <T> Codec<T> get(Class<T> clazz) {
...
   }
 
+  /**
+   * Appends a TypeAdapterFactory to allow Gson and Bson adapters to co-exist.
+   * Avoids registering the TypeAdapterFactory if Gson already has factories for BSON types.
+   *
+   * @param gson preconfigured instance
+   * @return {@code gson} configured with delegating type adapter factory.
+   */
+  public static Gson newGsonWithDelegatingTypeAdapterFactory(Gson gson) {
```
asereda-gs, line 162: Maybe call it `newGsonWithMongoSupport` or `withBsonSupport(Gson)` or `addBsonSupport(Gson)`.

`WithDelegatingTypeAdapterFactory` doesn't fully convey the meaning of this method.

mongo/src/org/immutables/mongo/repository/RepositorySetup.java
```diff
@@ -128,7 +127,7 @@ public String translateName(Member member) {
     class GsonNamingStrategy implements FieldNamingStrategy {
```
asereda-gs, line 127: I would move this class one level up it is already inside `RepositorySetup.FieldNamingStrategy`.

On the other hand, `FieldNamingStrategy` is already Single Abstract Method (SMA) interface so having a lambda on the top of `Gson` instance should be trivial for clients of this API.

Up to you ...

mongo/src/org/immutables/mongo/bson4gson/GsonCodecs.java
elucash: right, it's more important what the purpose of the action is (adds bson support) is than in which way it's technically achieved (by delegating type adapter factory)

mongo/src/org/immutables/mongo/repository/RepositorySetup.java
elucash: I've been thinking about it for a short while. Seems that we better have something like `GsonNamingStrategy` public (or private with public factory method `namingStrategyFrom(Gson)`) because it sort of documents that it's possible and should be used sometimes. It is easy implement with lambda, but people might have no clue that they ought to do that and do smth hardcoded.

asereda-gs: Perhaps `GsonNamingStrategy` warrants its own file and be public or, at least, `gsonNamingStrategy` factory method (also public).

Not 100% sure if it should be inside `bson4gson` package / `GsonCodes` class.

elucash: Maybe, but given that we think criteria is a newer implementation and to minimize impact of changes (as it is the goal of some changes asked for this PR) let it be just public `GsonNamingStrategy`, just can be moved out from interface and be directly nested in `RepositorySetup`  as `RepositorySetup.GsonNamingStrategy`. I think this will be minimal and cute enough.
P.S. I'll answer on other comments tomorrow morning.

mongo/src/org/immutables/mongo/bson4gson/GsonCodecs.java
jmoghisi: addressed by renaming it to `newGsonWithBsonSupport`

mongo/src/org/immutables/mongo/repository/RepositorySetup.java
jmoghisi: made it `public` visibility and moved to `RepositorySetup`

mongo/test/org/immutables/mongo/bson4gson/GsonCodecsTest.java
```diff
@@ -22,7 +25,7 @@ public void reflectiveTypeAdapter() {
   }
 
   @Test
-  public void dateCodec() throws IOException {
+  public void typeAdapterFromCodec() throws IOException {
```
asereda-gs, line 28: I think keeping the old name (`dateCodec`) makes more sense

mongo/test/org/immutables/mongo/bson4gson/GsonCodecsTest.java
```diff
@@ -36,4 +39,16 @@ public void dateCodec() throws IOException {
     check(doc.get("$date").getBsonType()).is(BsonType.DATE_TIME);
     check(doc.get("$date").asDateTime().getValue()).is(date.getTime());
   }
+
+  @Test
+  public void newGsonWithDelegatingTypeAdapterFactory() {
```
asereda-gs, line 44: Maybe call it `gsonWithBsonSupport`

asereda-gs (approved):
Overall looks fine to me.

Just small comments before merging.

mongo/test/org/immutables/mongo/bson4gson/GsonCodecsTest.java
jmoghisi: addressed

mongo/test/org/immutables/mongo/bson4gson/GsonCodecsTest.java
jmoghisi: I went with `typeAdapterFromDateCodec`

elucash:
Thank you! I think this is good and useful! I'm merging this. The release will be when I'll have my broadband internet restored, should be soon, but no ETA

elucash:
I'll also squash this (I'm forgetting sometimes to squash other PRs, sorry), so you'll see one commit in the master.

jmoghisi:
Thanks Eugene, appreciate all the feedback as well.

Issue #1408, opened 2022-10-10
Feature request: Generate update(UnaryOperator)-methods along withers (new style flag?)


ajkettun:
Currently transforming Immutables-generated objects requires to write rather clumsy code.
```
immutableBar.withFoo(doSomeTransformation(immutableBar.getFoo()))
```
And it gets worse with nested objects.

Simple improvement could be if Immutables would generate update-methods along with withers to allow transforming nested objects more cleanly. Basic update-method signature for a `String foo`-property in ImmutableBar -class would be:
```
ImmutableBar updateFoo(UnaryFunction<String> updater);
```
Which would allow to write:
```
immutableBar.updateFoo(foo -> ...)
```
Or if the `foo`-property would be another immutables -generated object instead of String:
```
immutableBar.updateFoo(foo -> foo.updateSomeOtherField(someOtherField -> ... ))
```
Currently updaters can be written by hand to the template class / interface if needed, but I see no reason why they could not be generated.

Also Optional (and Vavr Option etc.) could be taken account and offer an updater method which would transform the value only if it is present but this feature is not a must.

Chaining and doing this for multiple fields is of course inefficient but nothing prevents abusing withers in the same way.

ajkettun:
That was fast! Thank you! 

Hopefully records get withers some day and they creep up to regular immutable classes too:

https://github.com/openjdk/amber-docs/blob/master/eg-drafts/reconstruction-records-and-classes.md

elucash:
yep, all good stuff comes to java only if you don't wait for it ;)

Issue #590, opened 2017-04-03
Modifying deep immutable graphs (code verbosity)


asereda:
Hello, 

We have a usecase when we would like to modify small number of attributes in a deeply nested  immutable object (structure). Is there an **easy** way to do so, without manually wrapping / unwrapping it into Builder / [Modifiable](http://immutables.github.io/immutable.html#modifiable-classes)s  ?

[deepImmutablesDetection ](https://github.com/immutables/immutables/blob/master/value-fixture/src/org/immutables/fixture/deep/Canvas.java) helps to some extent but, unfortunately, verbosity of the code doesn't disappear ?

## Example
Simple Parent / Child relationship : 
```java
interface Parent {
  List<Child> children();
}

interface Child {
   long id();
   String name();
}
```
Say I want to change name of Child (5) to John the following operations have to be performed : 
```java
Parent parent = ... // get parent
final Child child = ImmutableChild.of(5, "John");
final List<Child> children = new ArrayList(parent.children()); // original list is also immutable
children.replaceAll( c -> c.id() == child.id() ? child : c);
parent = ImmutableParent.builder().from(parent).children(children).build();
```
Things get even more complicated when one has deeper nesting like Entity / Person  / Child etc.

Thanks in advance,
Andrei.

elucash:
Are you talking about something like Lenses/Prisms/Traversables? I had a plans to create something like that for the idiomatic Java and Immutables, but it requires some more experimentation to flesh out the concept.
But there's something else I've used for similar use cases: generation of visitor-like tree transformers. It poses its own limitations on how you'd define your model, but it worked fine enough. I'd also say that it doesn't remove much of the code, but can make it more manageable. As it's not documented just the starting point could be looking at https://github.com/immutables/immutables/blob/master/trees/test/org/immutables/trees/ast/SampleNodes.java and see SampleNodesTransformer (need to run build) generated using `@Trees.Transform` annotations.

asereda:
Hi Eugene,
Yes I was thinking something like Lenses / Traversables (although I'm not a haskell expert).

Is there any example / documentation of how to use generated Transformer (SampleNodesTransformer) ?  Or how to use (immutable) trees in general.

elucash:
0. No docs, sorry
1. Create some types, possibly including subtype hierarchies, within `@Value.Enclosing` umbrella type, put `@Trees.Transform` from `org.immutables:trees` module
2. `[UmbrellaClassName]Transformer` is generated.
[...]

PR #1409, opened 2022-10-12
#1408 withUnaryOperator style flag/naming template

elucash:
Implementation of #1408 using naming template as enabling style flag

### DIFF
--- a/value-fixture/test/org/immutables/fixture/with/WithUnaryOperatorTest.java
+++ b/value-fixture/test/org/immutables/fixture/with/WithUnaryOperatorTest.java
@@ -1,0 +1,43 @@
+package org.immutables.fixture.with;
+
+import java.util.Optional;
+import org.junit.jupiter.api.Test;
+import static org.immutables.check.Checkers.check;
+
+public class WithUnaryOperatorTest {
+	@Test
+	public void mapAttributes() {
+		ImmutableWithUnaryOperator m = ImmutableWithUnaryOperator.builder()
+				.a(1)
+				.b("b")
+				.addL("x", "y", "z")
+				.o("**")
+				.build();
+
+		ImmutableWithUnaryOperator t = m.mapA(a -> a + 1)
+				.mapB(b -> "[" + b + "]")
+				.mapL(String::toUpperCase)
+				.mapO(o -> o.replace('*', '_'));
+
+		check(t.a()).is(2);
+		check(t.b()).is("[b]");
+		check(t.l()).isOf("X", "Y", "Z");
+		check(t.o()).is(Optional.of("__"));
+	}
+
+	@Test
+	public void mapEmptyReturningThis() {
+		ImmutableWithUnaryOperator m = ImmutableWithUnaryOperator.builder()
+				.a(1)
+				.b("b")
+				.build();
+
+		ImmutableWithUnaryOperator t = m.mapO(o -> o + "!!!")
+				.mapL(l -> l + "???")
+				.mapI(i -> ++i)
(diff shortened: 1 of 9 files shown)


elucash:
walk-in reviews are welcome

value-fixture/test/org/immutables/fixture/with/WithUnaryOperatorTest.java
```diff
@@ -0,0 +1,43 @@
...
+        .a(1)
+        .b("b")
+        .addL("x", "y", "z")
+        .o("**")
+        .build();
+
+    ImmutableWithUnaryOperator t = m.mapA(a -> a + 1)
+        .mapB(b -> "[" + b + "]")
+        .mapL(String::toUpperCase)
+        .mapO(o -> o.replace('*', '_'));
```
asereda-gs, line 20: does it make sense to have `flatMap` on optionals ?

asereda-gs:
Eugene, a while ago you mentioned a more generic and functional approach to modifying Immutables: Lenses/Prisms/Traversables (see [comment](https://github.com/immutables/immutables/issues/590#issuecomment-291293621)). Is it still something you're considering ?

value-fixture/test/org/immutables/fixture/with/WithUnaryOperatorTest.java
elucash: Maybe, idk. Because how this feature is activated (by a naming template), I see that flatmap will require separate style flag to enable, `Style.withFlatmap` for example.
Also it would be fair for collections to  also have flatmap, but it will probably need 2 overloads,  like  `Function<? super T,  ? extends Iterable<? extends T>>` and `Function<? super T,  ? extends Stream<? extends T>>`.
BTW (DYK)  I've just learned couple of days ago that you can return `null` in the map function (JDK Optional)

```java
assert Optional.of("X").map(x -> null).isEmpty(); // no NPE
```

I think if there would be any demand for flatMap, which I see as a separate flag, I can implement flatMap for optional / collections. Otherwise, I think, Encodings should be used to customize any other specific container or collection.
I've already planned to add smth simple for `map`, there were at least couple of use cases where this is useful for us. Flatmap is more exotic in that sense, that while it's logical to have, we have never needed it on the surface of immutable object.

elucash:
> Eugene, a while ago you mentioned a more generic and functional approach to modifying Immutables: Lenses/Prisms/Traversables (see [comment](https://github.com/immutables/immutables/issues/590#issuecomment-291293621)). Is it something you're still considering ?

Yes, that still on the radar somewhere. Here, I've tried to address only simple a tool for mapping. Right now I'm not sure that this particular usecase would be simpler if lenses are used, it would be more scary looking for sure. But that is actually the API design exercise. I've got to designing something similar while working on Typescript codegen for the data models, the result was pretty interesting and not exactly what I thought it would be, some food for thought. So, the question is if there is demand for this? 

The other thing on the radar is the idea of Immutables 3 with more manageable models and many plugin points, this is for some newer LTS (JDK 11+ or even JDK 17+). And any complex new functionality could be implemented on top of that.

elucash:
thanks for the comments! I think if there's strong demand for related functionality we should discuss it further. Will merge this one. I wanted it to be a simple "no ambition" feature to update things with the function and not be on par with functor-ish (or monadic) `map`/`flatMap`. Like in React, you update state either with a value or a function which computes value from the previous value and that's what this is.
Albeit we want to have a separate, strong framework for updating immutable trees, a kind of "optics", but more java-esque (than haskell-like)

Issue #1407, opened 2022-10-07
Staged builder breaks jackson deserialization of nullable fields


SimY4:
Consider an entity definition:

```
@Value.Immutable
@Value.Style(stagedBuilder = true)
@JsonDeserialize(as = ImmutableEntity.class)
public interface Entity {
  @Value.Default
  default String getNullableDefault() {
    return "default";
  }
}
```

and a JSON string: `{"nullableDefault":null}`

This would fail jackson deserialisation with:

```
Can not construct instance of org.immutables.fixture.jackson.ImmutableEntity, problem: nullableDefault
 at [Source: {"nullableDefault":null}; line: 1, column: 40]
	at org.immutables.fixture.jackson.ObjectMappedTest.jsonNullableWithDefault(ObjectMappedTest.java:252)
Caused by: java.lang.NullPointerException: nullableDefault
	at org.immutables.fixture.jackson.ObjectMappedTest.jsonNullableWithDefault(ObjectMappedTest.java:252)
```

If you remove `@Value.Style` serialization works as expected.

elucash:
Thank you for reporting this. It's hard to dig every feature combination, especially with quite complex Jackson mapping. Here I would ask for help, If staged builder is involved then how it exactly affects Jackson so it doesn't work? One can, for example, take the generated code and out it into a source folder, remove `@Value.Immutable` annotation and try to fiddle / debug the code to find out what causes the problem.

PR #1410, opened 2022-10-16
fixing #1407 - Default value with explicit null should use default.

SimY4:
Fixing an issue with Jackson and staged builder feature interaction.

Essentially, null values are not allowed when the immutable entry is identified as not nullable.


elucash:
I'm not sure how Jackson/Immutables combination work in this case in terms of semantics, It's hard to me to say that it is justified of not to behave in this way, also I'm not sure how it's only surface with staged builder, seems to be a common problem. However, it's still looks innocent, even if a little bit lenient/non-strict behavior, so I'm merging this. I have no better way to fix this, and I have not strong opinions (any more) about how Jackson/Immutables combination should work in ideal word. Thank you for the PR!

Issue #1376, opened 2022-04-02
criteria.backend.WriteResult returns null | MongoDB


shahar080:
Hey,

- Using MongoDB and Criteria
When trying to perform a add/update/delete operation, the WriteResult has 3 fields with null value, doesn't matter if the operation was successful or not.

Example:
![image](https://user-images.githubusercontent.com/30349852/161399740-f651b4ee-8817-4932-87d3-d320292f8934.png)

shahar080:
Any updates regarding this?

dparamoshkin:
Have the same issue:(

dparamoshkin:
```  private <T> Publisher<WriteResult> update(StandardOperations.Update operation) {
    ReplaceOptions options = new ReplaceOptions();
    if (operation.upsert()) {
      options.upsert(operation.upsert());
    }

    List<ReplaceOneModel<Object>> docs =  operation.values().stream()
            .map(value -> new ReplaceOneModel<>(new BsonDocument(Mongos.ID_FIELD_NAME, toBsonValue(keyExtractor.extract(value))), value, options))
            .collect(Collectors.toList());

    Publisher<BulkWriteResult> publisher = ((MongoCollection<Object>) collection).bulkWrite(docs);
    return Flowable.fromPublisher(publisher).map(x -> WriteResult.unknown());
  }```

looks like some shortcuts was done in `return Flowable.fromPublisher(publisher).map(x -> WriteResult.unknown());`

asereda-gs:
This should be fixed. Please let me know if it works for you.

dparamoshkin:
@asereda-gs thank you for the fix, unfortunately mongo itself have quite weird behavior and not counting some cases correctly.

so if I use `upsert`  for new entry mongo returns (so insertCount is 0 but upsert collection have 1 element in it)
![image](https://user-images.githubusercontent.com/1911166/197528589-fb6b9b50-6cc5-47ea-ba8e-cbdd192a5209.png)
when I use `upsert' again for updating entry, it's working as expected (updateCount = 1)
![image](https://user-images.githubusercontent.com/1911166/197528554-29c8b914-1578-45e9-9ae8-117bfeee2a9f.png)

PR #1412, opened 2022-10-24
Fix counts in WriteResult for upsert operation

dparamoshkin:
This PR fixes insert counters for the upsert operation described in https://github.com/immutables/immutables/issues/1376  Mongo doesn't provide it as `insertedCount`, but it could be get from the size of `upserts` collection


asereda-gs:
Thanks for this PR!

asereda-gs:
Now when I think, perhaps `WriteResultTest` can be shared among multiple backends.

Take a look at [WriteTemplate](https://github.com/immutables/immutables/blob/master/criteria/common/test/org/immutables/criteria/typemodel/WriteTemplate.java). Maybe something similar can be created (eg. called `WriteResultTemplate`) and applied to all backends which support returning write counts ?

@dparamoshkin is it something you might be interested to implement ?

asereda-gs:
Or another option is to enhance `WriteTemplate` by checking `WriteResult` of different operations (if counts are present).

dparamoshkin:
@asereda-gs I would check on weekend if I can improve something there

Issue #1387, opened 2022-05-22
Criteria Reader doesn't respect getter prefixes


andonimichael:
My datamodels are setup as interfaces with a style of `get = {"get*", "is*"}`. The Criteria repository respects these prefixes when writing the data to Mongo, but does not respect them when reading the data. Specifically, the fields are written as "x", but the query references them as "getX", causing the fetch to fail.

----

My datamodels are of the shape:

```
// Only relevant style kept for brevity
@Style(get = {"get*", "is*"},
    implementationNestedInBuilder = true,
    init = "set*",
    overshadowImplementation = true)
@Criteria
@Immutable
@JsonSerialize(as = ImmutableMyDataModel.class)
@JsonDeserialize(builder = MyDataModelBuilder.class)
@Repository(facets = {SyncReadable.class, SyncWritable.class})
public interface MyDataModel {
    @Auxiliary
    @Criteria.Id
    @Derived
    @JsonProperty("_id")
    default String getId() {
        return randomUUID().toString();
    }

    String getExampleData();
}
```

(Note I have verified this issue with both Sync and Async facets. I have not tested Rx facets)

I see the data correctly written in Mongo as:
```
_id: "ab1a393f-0143-445c-af43-5ade8525c52d"
exampleData: "Test"
```

However, the read fails to return anything when using criteria filtering. I first verified this via the queries from my mongo logs:

```
"filter" : {
        "$and" : [
	    {
	        "getExampleData" : {
	            "$eq" : "Test"
	        }
	    }, ...
	]
}
```

And then later traced it down to the path fields in the CriteriaContext#state being incorrectly set to "getX".

----

I feel like two things are broken here:
1. There is an inconsistency between the Writer and Reader. I am fine if both write/read as getX or (preferably) both write/read as x; but having them not use the same field naming creates a very fragile design that pushes the synchronization up to the user.
2. There seem to be an inconsistency between the supported options in the Immutables.Value and the Immutables.Criteria frameworks. Ideally, all supported configuration in Value is also naturally supported in Criteria.

Some clarification in the web documentation might be helpful here too.

Thanks,
_Andoni_

andonimichael:
Any update? The incompatibility in the serializer and deserializer seems like a legitimate bug that is directly affecting one of my projects.

elucash:
[...]

PR #1417, opened 2022-12-01
1387 criteria reader doesn't respect getter prefixes

SpyrosKou:
This is an attempt to fix autogenerated Criteria that do not work in `find()` in the `MongoBackend` an issue also described in https://github.com/immutables/immutables/issues/1387 with more details.
The conversion from the property names to the MongoDB fields seems to take place in `MongoSession.java`.
There the `_id` property was handled individually and all other properties were delegated to `PathNaming.defaultNaming()` which returned the path as a field name.  This did not work in MongoDB filter queries. As for example it returned `getDescription` instead of `description` and the filter query required the field name, being `description`.
Replacing the `PathNaming.defaultNaming()` with the `JavaBeanNaming()` provides accurate field name transitions.

2 tests have also been included that test that `find() `works with the `ID`, with a getter of the form `getProperty()' and with a getter of the form `property()'.  I named properties more specifically to facilitate any potential future references.

PS: This PR solves some cases and improves the situation, but more changes should be examined. For instance taking into account the `@JsonProperty` annotation.

### DIFF
--- a/criteria/mongo/src/org/immutables/criteria/mongo/MongoSession.java
+++ b/criteria/mongo/src/org/immutables/criteria/mongo/MongoSession.java
@@ -38,17 +38,7 @@ import org.bson.codecs.EncoderContext;
 import org.bson.codecs.configuration.CodecRegistries;
 import org.bson.codecs.configuration.CodecRegistry;
 import org.bson.conversions.Bson;
-import org.immutables.criteria.backend.Backend;
-import org.immutables.criteria.backend.BackendException;
-import org.immutables.criteria.backend.DefaultResult;
-import org.immutables.criteria.backend.ExpressionNaming;
-import org.immutables.criteria.backend.KeyExtractor;
-import org.immutables.criteria.backend.PathNaming;
-import org.immutables.criteria.backend.ProjectedTuple;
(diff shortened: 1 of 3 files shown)


elucash:
@asereda-gs can you check this when you have time. I would just merge it, but need to understand how well it addresses the underlying problem and if it causes any issues / incompatibilities

asereda-gs:
@elucash let me check this PR on weekend. Sorry was busy lately.

asereda-gs (commented):
Hello.

Thank you for this PR!

I'm thinking if allowing users to register their own `PathNaming` instance in `MongoSetup` would be a more flexible strategy?


Overall `PathNaming` needs to be better integrated with Gson/Jackson/Bson serializations. 

I assume this issue also affects ElasticSearch backend?

asereda-gs:
Ideally one should re-use existing  BSON `Codec` which is being employed when entities are being persisted. The problem is that we can't know the path mapping unless the entity is serialized.

asereda-gs:
I have prepared some ideas in a separate branch: https://github.com/immutables/immutables/tree/jackson-pathnaming

it needs some more work

asereda-gs:
Please take a look at 3fb8e3 and let me know if it works for you.

PR #1441, opened 2023-02-15
Update link in README

RajatArora08:


elucash:
Thank you for PR, sorry, but it's actually `http://` not `https://`. For my browser it's not even loading from `https://`. Currently, the domain only serves a redirect to `https://immutables.github.io/`.

Issue #1442, opened 2023-02-17
Criteria for Mongo returns wrong results when sorting by id property


harmenweber:
When using criteria for mongo, sorting by the id property returns incorrect results. More precisely, the results are not sorted by the id property even though you specified a sorting by id.

Here is a concrete example.

Given the interface `StringHolder`:

```java
  @Value.Immutable
  @Criteria
  @Criteria.Repository
  @JsonDeserialize(as = ImmutableStringHolder.class)
  @JsonSerialize(as = ImmutableStringHolder.class)
  @JsonIgnoreProperties(ignoreUnknown = true)
  interface StringHolder {
    @Criteria.Id
    String id();
    String value();
    @Nullable String nullable();
    Optional<String> optional();
    String[] array();
    List<String> list();
  }
```

When you insert a bunch of `StringHolder` instances into a collection and then fetch all ordered by id, you fill receive all `StringHolder` instances in an incorrect order. Let's be fair, there is a small chance that you get them in the correct order, but that would be by accident.

The following test fails:

```java
  @Test
  void sortById() {
    final StringHolderRepository repository = new StringHolderRepository(this.resource.backend());

    final List<TypeHolder.StringHolder> stringHolders =
        Stream.generate(TypeHolder.StringHolder.generator()).limit(100).collect(Collectors.toList());
    Collections.shuffle(stringHolders);
    repository.insertAll(stringHolders);

    final List<String> expectedIdsInOrder =
        stringHolders.stream().map(TypeHolder.StringHolder::id).sorted().collect(Collectors.toList());

    final List<String> actualIds = repository.findAll().orderBy(StringHolderCriteria.stringHolder.id.asc()).fetch()
        .stream().map(TypeHolder.StringHolder::id).collect(Collectors.toList());

    Checkers.check(actualIds).is(expectedIdsInOrder);
  }
```

The reason is, that the `MongoSession` does not apply the `PathNaming` when handling the collations in the query. As a result, the id property does not get translated to Mongo's `_id` field. Using the `PathNaming` when handling the collations will solve the problem. The `PathNaming` is aware of the id property and will translate it to `_id`.

harmenweber:
Here is a pull request to fix that problem: #1443

asereda-gs:
Fixed in #1443

PR #1443, opened 2023-02-17
Fix sorting by id property with Criteria for Mongo #1442

harmenweber:


elucash:
Thank you for the PR! The fix is minimal, tested and seems on point. @asereda-gs if there's no objection I'll merge it. I do realize that similar issues can plague other places in criteria, but it would be out of scope to investigate that at the moment.

asereda-gs:
I'm fine merging this PR as is.

Probably next step would be to create a separate Template test class for sorting (similar to `BooleanTemplate`).

elucash:
Thank you!

Issue #1450, opened 2023-03-22
Exception when compiling large projects in Eclipse


hwellmann:
Not a bug ticket, but a hint for others running into the same issue:

### Problem

We have a project with about 3000 Java source files some of which use `Immutables.Value`. The project compiles with Maven on the command line, even when using the Eclipse compiler (`<compilerId>eclipse</compilerId>` in `maven-compiler-plugin` configuration), but it fails to compile in the Eclipse IDE.

There are a stacktraces of the following kind in the Error Log:

```
org.immutables.value.internal.$processor$.$Processor threw java.lang.NullPointerException: Cannot read field "referenceContext" because "this.scope" is null
	at org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding.sourceStart(SourceTypeBinding.java:3272)
	at org.eclipse.jdt.internal.compiler.lookup.MethodBinding.sourceStart(MethodBinding.java:1298)
	at org.eclipse.jdt.internal.compiler.apt.model.TypeElementImpl$SourceLocationComparator.determineSourceStart(TypeElementImpl.java:104)
	at org.eclipse.jdt.internal.compiler.apt.model.TypeElementImpl$SourceLocationComparator.getSourceStart(TypeElementImpl.java:76)
	at org.eclipse.jdt.internal.compiler.apt.model.TypeElementImpl$SourceLocationComparator.compare(TypeElementImpl.java:69)
	at org.eclipse.jdt.internal.compiler.apt.model.TypeElementImpl$SourceLocationComparator.compare(TypeElementImpl.java:1)
	at java.base/java.util.TimSort.countRunAndMakeAscending(TimSort.java:355)
	at java.base/java.util.TimSort.sort(TimSort.java:220)
	at java.base/java.util.Arrays.sort(Arrays.java:1307)
	at java.base/java.util.ArrayList.sort(ArrayList.java:1721)
	at java.base/java.util.Collections.sort(Collections.java:179)
	at org.eclipse.jdt.internal.compiler.apt.model.TypeElementImpl.getEnclosedElements(TypeElementImpl.java:178)
	at org.immutables.value.internal.$generator$.$SourceOrdering$EclipseCompilerOrderingProvider.enclosedBy($SourceOrdering.java:126)
	at org.immutables.value.internal.$generator$.$SourceOrdering.enclosedBy($SourceOrdering.java:99)
[...]

PR #1452, opened 2023-03-22
detect the Eclipse compiler without using the TCCL

### LINKED ISSUE #1451: Support Java Language Server
Currently, projects using `Value.Immutable` cannot be compiled in Visual Studio Code or other environments based on the [Eclipse JDT Language Server](https://github.com/eclipse/eclipse.jdt.ls).

The workarounds implemented in Immutables for the Eclipse Compiler do not get activated in the Language Server environment.

hwellmann:
Fixes #1451.

The current implementation depends on the thread context class loader being set to `org.eclipse.osgi.internal.framework.ContextFinder` which is true in Eclipse IDE, but not in the JDT Language Server used by Visual Studio Code.

In addition, it seems that the `EclipseCompilerOrderingProvider` is no longer required. It can be disabled by defining the system property `-Dorg.immutables.disableEclipseOrderingProvider=true` in the VS code setting `java.jdt.ls.vmargs`.

(The system property and the `EclipseCompilerOrderingProvider` should both be removed if this is confirmed.)

Moreover, `java.jdt.ls.vmargs` should include `-DmaxCompiledUnitsAtOnce=10000` for large projects (see #1450).

### DIFF
--- a/generator/src/org/immutables/generator/SourceOrdering.java
+++ b/generator/src/org/immutables/generator/SourceOrdering.java
@@ -67,6 +67,9 @@ import org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding;
  * </ul>
  * <p>
  * <em>Based on a workaround idea provided by Christian Humer</em>
+ * 
+ * NOTE: The Eclipse bug is fixed, and this workaround can be disabled by setting the system property
+ * {@code org.immutables.disableEclipseOrderingProvider=true}.
  */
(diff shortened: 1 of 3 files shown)


elucash:
Fully agree. If this really works than we should merge it :)

elucash:
I'll merge it and will try to see if still works in Eclipse. Hope someone will re-validate it works for the language server

PR #1468, opened 2023-06-07
Populate correct initial capacity for List/Set/Map/ImmutableList.Builder

snazy:
* Uses the `new ArrayList(int)` constructor when `size` is available
* Fixes the initial capacity calculation considering the default load factor of `.75` for sets + maps
* Uses `ImmutableList.builderWithExpectedSize` where appropriate

This change avoids a bunch of "resize" operations and some unnecessary object allocations / heap pressure.


elucash:
Thank you for the PR! it looks good, I just need another detailed look before merge

snazy:
@elucash You ok with the PR.
Would love to have this change being released, because that really brings a huge win, especially when pre-populated collections are passed to an immutable's `of` function.

elucash:
Thank you for remainder. I did have another look and it ok, merging it

Issue #291, opened 2016-03-31
unknown enum constant ImplementationVisibility.PACKAGE if style is applied to class directly


qinfchen:
The consuming project would have the following warning if the immutable class that the project is consuming has Style annotation directly applied to the class. This error disappears if i create a style interface and apply it to the class.
```
warning: unknown enum constant ImplementationVisibility.PACKAGE
  reason: class file for org.immutables.value.Value$Style$ImplementationVisibility not found
1 warning
```

qinfchen:
see https://github.com/palantir/dropwizard-web-security/pull/55

elucash:
I guess I know what is happening here, I also checked the build file of the project you've referenced and it's seems that you need to actually add annotation jar. The combined jar (value-x.x.x.jar) works alone when it's used on compile classpath. If you put processors on the separate class-path like `processor`, then you still need annotations to be in a compile classpath.
You can get some hints from the Gradle example for Android http://immutables.github.io/getstarted.html#android

For your configuration it would be something like this
```groovy
  processor "org.immutables:value:$immutablesVersion"
  compile "org.immutables:value:$immutablesVersion:annotations"
```
Hope it will solve the problem, thank you!

qinfchen:
thanks for the response. But I don't understand why the warning would disappear if i declare a new annotation with the same immutable styles.

elucash:
I have no precise explanation, but there are all sort of quirks in the implementation of annotation processing and interactions with the compiler for both Javac and ECJ. Warning might disappear but the annotation configuration might not get processed correctly. 

Anyway when you run the annotation processor in a separate scope/classpath, then you should have annotation api (including enums etc) on the regular compilation classpath.

qinfchen:
@elucash,  we don't want pull in a specific version of immutables, so that there is no version conflict for the consuming app. Do `annotations` change a lot?

elucash:
It's the same version, but the artifact have an `annotations` classifier which usually indicates different packaging. The classifier logically is the part of the artifact name rather than version qualifier. in our case it is the same `value` artifact but with only annotation included and no annotation processor classes. You put this artifact on the compilation classpath as you would do for any other dependency. You dont need neither annotations-only nor the annotation processor artifact in runtime, so you can use `compileonly` or `provided` sort of dependency scope.
[...]

PR #1474, opened 2023-07-28
Allow specifying visibility as string

nastra:
In https://github.com/apache/iceberg/pull/8099 we would like to use a custom style annotation to modify the name of the generated enclosing class (mainly to keep API compatibility).

```
@Value.Enclosing
@Value.Style(typeImmutableEnclosing = "ImmutableSnapshotTable")
interface BaseSnapshotTable extends SnapshotTable {

  @Value.Immutable
  interface Result extends SnapshotTable.Result {}
}
```
Now we'd also like to configure `visibility = Value.Style.ImplementationVisibility.PUBLIC` so that the generated class is publicly visible. 

However, doing that would mean that we'd be forcing a compile-time dependency to consumers due to the issue described in https://github.com/immutables/immutables/issues/291 and the currently suggested workaround for this is to have something like
```
@Retention(RetentionPolicy.SOURCE)
@Value.Style(visibility = Value.Style.ImplementationVisibility.PUBLIC)
@interface ImmutablesPublic {}
```

But applying both `@Value.Style(typeImmutableEnclosing = "ImmutableSnapshotTable")` and `@ImmutablesPublic` doesn't work due to applying two Style definitions on the class. 

One could argue to move `@Value.Style(typeImmutableEnclosing = "ImmutableSnapshotTable")` into `@ImmutablesPublic` but it's not feasible to create X custom meta annotations.

Hence my suggestion would be to allow specifying the visibility as a string in order to avoid the issue described in #291 and to not have to introduce custom meta annotations.

@elucash could you take a look at this please?

### DIFF
--- a/value-annotations/src/org/immutables/value/Value.java
+++ b/value-annotations/src/org/immutables/value/Value.java
@@ -917,6 +917,14 @@ public @interface Value {
     ImplementationVisibility visibility() default ImplementationVisibility.SAME;
 
     /**
+     * Specify the mode in which visibility of generated value type is derived from abstract value
+     * type. Specifying this will override {@link #visibility()} and the reason why this option as a string exists,
+     * is to avoid javac warnings mentioned in <a href="https://github.com/immutables/immutables/issues/291">#291</a>.
+     * @return implementation visibility as string
+     */
+    String visibilityString() default "";
+
+    /**
      * Specify whether init, copy and factory methods and constructors for an unwrapped {@code X} of
      * {@code Optional<X>}
      * should accept {@code null} values as empty value. By default nulls are rejected in favor of
@@ -1171,6 +1179,15 @@ public @interface Value {
      * @return implementation visibility
      */
     BuilderVisibility builderVisibility() default BuilderVisibility.PUBLIC;
+
+    /**
+     * Specify the mode in which visibility of generated value type is derived from abstract value
+     * type. Specifying this will override {@link #builderVisibility()} and the reason why this option as
+     * a string exists, is to avoid javac warnings mentioned in
+     * <a href="https://github.com/immutables/immutables/issues/291">#291</a>.
+     * @return implementation visibility as string
+     */
+    String builderVisibilityString() default "";
 
     /**
      * Runtime exception to throw when an immutable object is in an invalid state. I.e. when some
--- a/value-processor/src/org/immutables/value/processor/meta/Constitution.java
+++ b/value-processor/src/org/immutables/value/processor/meta/Constitution.java
@@ -39,6 +39,7 @@ import org.immutables.value.processor.meta.Proto.Protoclass;
 import org.immutables.value.processor.meta.Reporter.About;
 import org.immutables.value.processor.meta.Styles.PackageNaming;
 import org.immutables.value.processor.meta.Styles.UsingName.TypeNames;
+import org.immutables.value.processor.meta.ValueMirrors.Style.BuilderVisibility;
 import org.immutables.value.processor.meta.ValueMirrors.Style.ImplementationVisibility;
 import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.base.Verify.verify;
@@ -62,7 +63,14 @@ public abstract class Constitution {
 
 	@Value.Derived
 	public Visibility implementationVisibility() {
-		if (style().visibility() == ImplementationVisibility.PRIVATE
+    ImplementationVisibility visibility;
+    if (!style().visibilityString().isEmpty()) {
+      visibility = ImplementationVisibility.valueOf(style().visibilityString());
+    } else {
+      visibility = style().visibility();
+    }
+
+		if (visibility == ImplementationVisibility.PRIVATE
 				&& !protoclass().features().builder()
 				&& !protoclass().kind().isNested()) {
 			protoclass()
@@ -73,12 +81,17 @@ public abstract class Constitution {
 									"required");
 			return Visibility.PACKAGE;
 		}
-		return protoclass().visibility().forImplementation(style().visibility());
+		return protoclass().visibility().forImplementation(visibility);
 	}
 
 	@Value.Derived
 	public Visibility builderVisibility() {
-		Visibility visibility = protoclass().visibility().forBuilder(style().builderVisibility());
+    Visibility visibility;
+    if (!style().builderVisibilityString().isEmpty()) {
+      visibility = protoclass().visibility().forBuilder(BuilderVisibility.valueOf(style().builderVisibilityString()));
+    } else {
+      visibility = protoclass().visibility().forBuilder(style().builderVisibility());
+    }
 		if (visibility == Visibility.PUBLIC && protoclass().styles().style().stagedBuilder()
 				&& isNestedFactoryOrConstructor()) {
 			return Visibility.PRIVATE;
--- a/value-fixture/src/org/immutables/fixture/style/MoreVisibleImplementationAsString.java
+++ b/value-fixture/src/org/immutables/fixture/style/MoreVisibleImplementationAsString.java
@@ -1,0 +1,38 @@
+/*
+   Copyright 2014 Immutables Authors and Contributors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ */
+package org.immutables.fixture.style;
+
+import org.immutables.value.Value;
+import org.immutables.value.Value.Style.ImplementationVisibility;
+
+/**
+ * Feature combination
+ * <ul>
+ * <li>Implementation is public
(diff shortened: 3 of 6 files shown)


elucash:
Thank you for the PR. One one hand, all-in-all-in-the-end we might merge smth like this. On the other hand, the change is very questionable, and I want to make sure we understand context of a problem. I've reread linked tickets, so I've started to recall some of that, but anyway. First, we need to answer the following question:

* Given someone uses annotation, and that annotation is annotated with another annotation which also uses enum, coming from the same lib (annotation class and enum class nested). It is a compile time dependency, why it is somehow bad to have this dependency?

> However, doing that would mean that we'd be forcing a compile-time dependency to consumers due to the issue described

This quoted statement is not something I can agree on. This might be true and I might be missing something or just plain wrong. My current take is that, If you use some annotations to some effect you should have these annotation on the classpath. But I want to put it into context, some explanation, which is important:
1. In some cases, annotations don't have to be on the classpath, but that for the cases where you don't do anything with them. Example: you use library XYZ, this library used compile-only annotation library ABC for compile time checking of something. You don't use ABC, so don't have it your classpath and XYZ still works for you without ABC.
2. In case of Style as meta annotation, compile-time processor would read that metadata, why deny it (the processor) access to the classes (`Style`, and `Style$ImplementationVisibility` in this case etc) which actually describe the types? Yes, annotation processor can work on compiled metadata sometimes on the basis on just classnames/symbol-names, but that might give Javac warning.
3.  We're providing (and that was from the very beginning) annotation-only jar. There were some discussions about that it should have been the default, but because of the backward compatibility we will not change this in this major version anyway. So there's `org.immutables:value-annotations` (artifact `value-annotations`) and `org.immutables:value:annotations` (artifact `value`, classifier `annotations`), these 2 jars have the same classes. If someone uses your annotation (`@ImmutablesPublic` etc) and it's meta-annotated with annotations from Immutables, it's absolutely nothing wrong with having Immutables annotations and enums being transitive dependency of your jar which supplies custom annotation. It's actually a right thing to do. Because this jar should be compilation only, it should not propagate to any dependency of jar which is compiled with the help of `@ImmutablesPublic`/`@Immutables.Value{Style}`
4. Be mindful of which classpath you use. (If we for a moment forget about Java Module Systems' new processor-module-path) you might need to have annotation-only jar on either processor-path or regular compile classpath, depending how you setup your annotation processor.

nastra:
@elucash thanks for your detailed description. 

> However, doing that would mean that we'd be forcing a compile-time dependency to consumers due to the issue described...

Sorry in case this wasn't precise. What I meant here is based on the below error prone warning, which refers to #291.

```
warning: [ImmutablesStyle] Using an inline Immutables @Value.Style annotation or meta-annotation with non-SOURCE retention forces consumers to add a Immutables annotations to their compile classpath.Instead use a meta-annotation with SOURCE retention.See https://github.com/immutables/immutables/issues/291.
interface BaseSnapshotTable extends SnapshotTable {
^
    (see https://github.com/palantir/gradle-baseline#baseline-error-prone-checks)
```
Now using a meta-annotation as I described above isn't feasible for us.
My ideal solution in https://github.com/apache/iceberg/pull/8099 would be to have the below class, where the class itself has **package-private** access, but we'd like to make the generated **ImmutableSnapshotTable** `public`.
```
@Value.Enclosing
@Value.Style(
    typeImmutableEnclosing = "ImmutableSnapshotTable",
    visibility = Value.Style.ImplementationVisibility.PUBLIC)
interface BaseSnapshotTable extends SnapshotTable {

  @Value.Immutable
  interface Result extends SnapshotTable.Result {}
}
```

Now when I publish these changes and consume this in another project (let's call it **ProjectX**), where I'd like to use **ImmutableSnapshotTable.builder()...build()**, we're seeing the below javac warning:

```
warning: unknown enum constant ImplementationVisibility.PUBLIC
  reason: class file for org.immutables.value.Value$Style$ImplementationVisibility not found
```

This is the part that I don't understand about this warning. **ImmutableSnapshotTable** itself doesn't contain any usage of `ImplementationVisibility` and  **ImmutableSnapshotTable** is correctly generated as `public`.

In order to get rid of the warning in **ProjectX**, one can add a dependency to `org.immutables:value-annotations` in **ProjectX**, but this isn't really a viable option, because consuming projects should not have to add anything immutable-specific to make that warning go away.

elucash:
This got me to seriously think about, this puzzle. Maybe I'm fooling myself, but I think I'm starting to understand what's going on.
So here's the train of thoughts, before it's gone

1. We've made (or left) Immutable.Style annotations CLASS-retained because we want to load these across compiled libraries, so it's possible to have precompiled jar dependency of styles (meta-annotated etc). And because of that and anyone else depending on that retention for any other reason, we cannot change that (i.e. cannot make those SOURCE)
2. You would be ok to make your style-annotation as SOURCE retained, but you need those to be customized per class, so you would use custom name and style-annotation. But because we don't have any merging of styles (not in this version, potential future, Immutables 3 will probably allow this, as necessity for more extensible design/plugins). So this is the reason to resort to inline style, which (flagged by error-prone).
3. The Javac warning is generated for the following reason in the ProjectX which uses generated object: When loading generated class it also loads its supertypes as bytecode (during compilation) and examines annotations. Because style annotation sitting on the superinterface is class retained and it's there in the bytecode, it tries to examine it and fails (non-critically, as a warning) to load attribute with enum value. For some reason it's ok that annotation itself is not available on the classpath, but not ok if its attribute value's enum class is not. This warning, apparently, appeared somewhere after Java 9.
4. If I say that the solution in this PR is still on the table, then I'll would also like to see if there are other solutions to what you're want to achieve. As I've mentioned, style merging can be implemented, but for this major version it would be quite an undertaking. Separate, small, isolated solution like a new annotation for such custom naming is also something I'm thinking of. I've looked at the patterns in the naming, like shown in https://github.com/apache/iceberg/pull/8099  It seems like what you want is to have `BaseViewHistoryEntry` -> `ImmutableViewHistoryEntry` renaming. Now, there's seems to be a way to automate that. Tell me if this will not work for your case. So you have style attribute `typeAbstract`, it works in the opposite way of renaming. It detects the logical name of the value type by trimming suffix or prefix. It was specifically designed for users who wish to have `*Base`, `Abstract*`, and similar suffixes or prefixes in the abstract interfaces/classes. In this case, your shared style annotation might have `typeAbstract="Base*", typeImmutable="Immutable*"` , even skip `typeImmutable="Immutable*"` as it's the default.

value-annotations/src/org/immutables/value/Value.java
```diff
@@ -916,6 +916,14 @@
      */
     ImplementationVisibility visibility() default ImplementationVisibility.SAME;
 
+    /**
+     * Specify the mode in which visibility of generated value type is derived from abstract value
+     * type. It is a good idea to not specify such attributes inline with immutable values, but
+     * rather create style annotation (@see Style). Specifying this will override {@link #visibility()}.
```
elucash, line 922: Need to expand documentation with the note about Javac warning (and maybe error-prone's one) at least briefly

> It is a good idea to not specify such attributes inline with immutable values, but rather create style annotation (@see Style)

This sorta tries to explain this, but without context (of warnings) it's not explaining it well why these `*String()` exist

value-processor/src/org/immutables/value/processor/meta/Constitution.java
```diff
@@ -62,7 +63,14 @@ public Generics generics() {
 
 	@Value.Derived
 	public Visibility implementationVisibility() {
-		if (style().visibility() == ImplementationVisibility.PRIVATE
+		ImplementationVisibility visibility;
+		if (!Strings.isNullOrEmpty(style().visibilityString())) {
```
elucash, line 67: never should be null, better use `.isEmpty()`.  In general, usage of `isNullOrEmpty` and similar (on collections, strings) considered a code smell (even its javadoc suggest alternatives)

value-processor/src/org/immutables/value/processor/meta/Constitution.java
```diff
@@ -62,7 +63,14 @@ public Generics generics() {
...
+			visibility = ImplementationVisibility.valueOf(style().visibilityString());
+		} else {
+			visibility = style().visibility();
```
elucash, line 70: I like and use tabs for indents, but this repo uses 2 spaces indentation. (I've seen some other code accidentally has tabs, but that was sloppiness and exception to the rule)

elucash:
I think after thinking and realizing the "exceptional" role of what seems to be strange quirk (if not a bug) in Javac wrt class retained annotations, well I think we should merge this PR and provide this solution. This would be definitely a place where Immutables 3 (and I hope there will be such) should improve and just avoid getting into such cases. I've asked for some cosmetic changes and I'll merge the PR.
I still recommend checking if name detection will work so you can actually use reusable style annotation instead of inline styles, it will add consistency/reduce any typo errors in names put as string literals on top of each abstract value type.

value-processor/src/org/immutables/value/processor/meta/Constitution.java
nastra: makes sense, updated

value-processor/src/org/immutables/value/processor/meta/Constitution.java
mc-elukash: you're right, I'll reformat this file at some point

elucash: Yep, this file is in need for reformat, but thank you for getting your lines correctly

elucash:
Thank you @nastra for the PR and refinements, merged!

nastra:
Thanks @elucash for getting this in!

PR #1477, opened 2023-08-03
Re-try to resolve types in java.lang

### LINKED ISSUE #1475: Incorrect imports for generated code that clashes with `java.lang`
It seems that when a generated class has the same simple name as something in `java.lang`, it is not correctly imported into generated code in other packages. We discovered this with `Record` when targeting Java 17, and I've since observed this with both `Module` and `Record`.

I've pushed [an example branch](https://github.com/immutables/immutables/compare/master...HubSpot:immutables:module-classname).

<details>
  <summary>Example code for convenience</summary>

```java
package org.immutables.fixture.module;

@Value.Immutable
@Value.Style(typeAbstract = {"*IF"}, typeImmutable = "*")
public interface ModuleIF {
}

```

```java
package org.immutables.fixture.module.subpackage;

import org.immutables.fixture.module.Module;

@Value.Immutable
public interface HasModuleField {
  Module getModule();
}
```

</details>

Running `mvn clean test-compile` fails with the following:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project value-fixture: Compilation failure: Compilation failure:
[ERROR] /Users/sgutz/dev/immutables/value-fixture/target/generated-sources/annotations/org/immutables/fixture/module/subpackage/ImmutableHasModuleField.java:[28,13] error: ImmutableHasModuleField is not abstract and does not override abstract method getModule() in HasModuleField
[ERROR] /Users/sgutz/dev/immutables/value-fixture/target/generated-sources/annotations/org/immutables/fixture/module/subpackage/ImmutableHasModuleField.java:[39,16] error: getModule() in ImmutableHasModuleField cannot implement getModule() in HasModuleField
[ERROR]   return type java.lang.Module is not compatible with org.immutables.fixture.module.Module
[ERROR] /Users/sgutz/dev/immutables/value-fixture/target/generated-sources/annotations/org/immutables/fixture/module/subpackage/ImmutableHasModuleField.java:[38,2] error: method does not override or implement a method from a supertype
[ERROR] /Users/sgutz/dev/immutables/value-fixture/target/generated-sources/annotations/org/immutables/fixture/module/subpackage/ImmutableHasModuleField.java:[150,31] error: incompatible types: org.immutables.fixture.module.Module cannot be converted to java.lang.Module
[ERROR] -> [Help 1]
```

Looking at the generated code, you can see that `ImmutableHasModuleField` does not import `org.immutables.fixture.module.Module` as it should. Oddly, this did not seem to have any problems when `org.immutables.fixture.module.Module` was not generated with immutables.

PtrTeixeira:
Because `java.lang` types get automatically imported, they can be "resolved" without being resolved to the same thing that they are in the source file. This ultimately winds up preventing the generated code from compiling.

I'd love any guidance you're willing to give around this / around testing this out. Thank you!

Closes #1475

### DIFF
--- a/value-fixture/src-java-11/org/immutables/fixture/module/ModuleIF.java
+++ b/value-fixture/src-java-11/org/immutables/fixture/module/ModuleIF.java
@@ -1,0 +1,8 @@
+package org.immutables.fixture.module;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+@Value.Style(typeAbstract = {"*IF"}, typeImmutable = "*")
+public interface ModuleIF {
+}
--- a/value-fixture/src-java-11/org/immutables/fixture/module/subpackage/HasModuleField.java
+++ b/value-fixture/src-java-11/org/immutables/fixture/module/subpackage/HasModuleField.java
@@ -1,0 +1,11 @@
+package org.immutables.fixture.module.subpackage;
+
+import org.immutables.fixture.module.Module;
+import org.immutables.value.Value;
+
+
+@Value.Immutable
+@Value.Style(headerComments = true)
+public interface HasModuleField {
+        Module getModule();
+}
--- a/value-processor/src/org/immutables/value/processor/meta/TypeStringProvider.java
+++ b/value-processor/src/org/immutables/value/processor/meta/TypeStringProvider.java
@@ -174,6 +174,14 @@ class TypeStringProvider {
       }
 
       hasMaybeUnresolvedYetAfter |= importsResolver.unresolved;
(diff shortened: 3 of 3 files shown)


elucash:
Thank you for PR! This should be important to fix this. Let me some time to understand the problem, the solution space and the specific fix in this PR. The fix looks good and quite compact, just need to understand how it all works together.

elucash:
I'll merge it, but after investigation, might change the way it works, currently it's a bit enigmatic to me. Thank you!

PtrTeixeira:
Thank you!

PR #1492, opened 2023-11-27
Refactor

angel-christian25:


elucash:
sorry for the long delays. Because the changes are big it's reasonably hard to say if those are ok. I hoped someone else can look at it, so I don't have to learn this part of the codebase and remember many things from some years ago.
If there's still intention to land this, I can have another deeper look, given the following.
1. Formatting in many files are disregarded (some other style is introduced in some files, indentation etc, please match to the rest of the code)
2. Don't extract(move to top level) classes in `Naming`, no need to do so. Try to provide just minimal factory method if you need some of those and even consider avoiding it, is there really no other way?
3. It is unclear why move internals of `hasValueOrVirtualFields` to encoding? the new method is not used in this PR.
4. Which tests to rely on to verify that refactoring doesn't change the behavior?

PR #1496, opened 2023-12-01
Allows NaturalOrdering for Comparable type hierarchies

wjglerum:
Tried to write some of our code to use the `@Value.NaturalOrder` annotation for comparable wrapped types and it failed. A simple fix was to use an upper bound wildcard in the generated code. Also added a testcase with an example.


wjglerum:
@elucash what do you think?

elucash:
sorry for the very long delay. It was hard for me to understand this PR from the first look, seems like it's only allowing proper bounds for collections. Thank you for the PR, merging

wjglerum:
@elucash thanks again! When can we expect a new minor/patch release? Would you consider doing a small one this week with these small patches?

elucash:
version `2.10.1` released

PR #1502, opened 2024-01-25
Fix 'yield outside of switch expression'

alsin:
When using Immutables with Java versions supporting switch expressions (and the new `yield` keyword) and if one decides to define an immutable interface with say `getYield()` method:
```java
@Value.Immutable
public interface SomeInterface {
    int getYield();
}
```
the value processor generates an immutable class implementation which won't compile since the internal generated Builder class would have the following copying method:
```java
    @CanIgnoreReturnValue 
    public final Builder from(SomeInterface instance) {
      Objects.requireNonNull(instance, "instance");
      yield(instance.getYield()); <-- here the compiler will fail with 'Yield outside of switch expression'
      return this;
    }

```

It's enough to prepend the method call with `this.` and the compilation passes.

### DIFF
--- a/value-fixture/src/nonimmutables/YieldField.java
+++ b/value-fixture/src/nonimmutables/YieldField.java
@@ -1,0 +1,9 @@
+package nonimmutables;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface YieldField {
+
+    int getYield();
+}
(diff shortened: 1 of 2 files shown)


elucash:
Thank you for the PR, I wonder if there are other places affected.

alsin:
> Thank you for the PR, I wonder if there are other places affected.

You're welcome! Could be, I just came across this bug on this particular place. Had to fix it temporarily by changing the setter style (using `set*`).

Issue #924, opened 2019-03-18
Add `required` attribute to generated `@JsonProperty` on a field


JaroKuruc:
In order to better support JSON schema generation, it would be helpful if generated `@JsonProperty` on mandatory fields also contained `required=true` attribute. This would then be inline with the logic in generated builder which is checking that mandatory fields are set.

elucash:
thanks for pointing this out! while I'm not aware of much of the functional usage of the `JsonProperty.required` attribute, but having this attribute set should be good for consistency whether it is functional or just a metadata piece

PR #1515, opened 2024-05-14
Added support for the 'required' attribute for the generated JSON Property Annotation

harrismirza:
Hi,

Raising this PR to solve this issue raised a while back: https://github.com/immutables/immutables/issues/924.

Essentially when we generate a Value class, we are adding an 'JsonProperty' annotation, but this does not have the 'required=true' flag set for non-nullable fields.

This is useful when inspecting the classes for things like generating API specs and JSON Schemas.

I have set this to be controlled by a style flag, so users can disable this functionality if needed.

As seen in the docs (https://javadoc.io/doc/com.fasterxml.jackson.core/jackson-annotations/latest/com/fasterxml/jackson/annotation/JsonProperty.html#required--), this should not actually affect deserialization behaviour for most users of Immutables, as they would be using the builder for deserialisation.

### DIFF
--- a/value-annotations/src/org/immutables/value/Value.java
+++ b/value-annotations/src/org/immutables/value/Value.java
@@ -1143,6 +1143,8 @@ public @interface Value {
      */
     boolean forceJacksonPropertyNames() default true;
 
+    boolean setJacksonPropertyRequired() default true;
+
     /**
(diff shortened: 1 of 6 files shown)


elucash:
Thank you for the PR! This is good solution, merging

PR #1521, opened 2024-06-18
Use source retention for serial annotations

Xcelled:
I believe the intent here was that the Serial annotations not be retained in the compiled .class files, as the `@Serial` is annotated with Source retention and the annotations within are not copied to the generated immutable.

However, the retention meta-annotation does not apply to nested annotation definitions, so ultimately usages of `@Serial.Version` were being included in the generated classfile for the immutable _definition_, leading to problems finding the classfile if the Serial artifact was not on the classpath.

Here's a dump of the bytecode for an immutable definition with Serial.Version - you can clearly see the annotation is included by the compiler. (h/t @kmclarnon)

```
// class version 61.0 (61)
// access flags 0x601
public abstract interface com/hubspot/PersonIF {

  // compiled from: PersonIF.java

  @Lcom/hubspot/immutables/style/HubSpotStyle;() // invisible

  @Lorg/immutables/value/Value$Immutable;() // invisible

  @Lorg/immutables/serial/Serial$Version;(value=-22940920492L) // invisible
  // access flags 0x2609
  public static abstract INNERCLASS org/immutables/value/Value$Default org/immutables/value/Value Default
  // access flags 0x2609
  public static abstract INNERCLASS org/immutables/value/Value$Immutable org/immutables/value/Value Immutable
  // access flags 0x2609
  public static abstract INNERCLASS org/immutables/serial/Serial$Version org/immutables/serial/Serial Version
```

@suruuK @stevie400 @jaredstehler @PtrTeixeira @ggs5427

### DIFF
--- a/serial/src/org/immutables/serial/Serial.java
+++ b/serial/src/org/immutables/serial/Serial.java
@@ -39,6 +39,7 @@ public @interface Serial {
    * is not strictly needed if structural is present.
    */
   @Target({ElementType.TYPE, ElementType.PACKAGE, ElementType.ANNOTATION_TYPE})
+  @Retention(RetentionPolicy.SOURCE)
   public @interface Version {
     long value();
   }
@@ -59,5 +60,6 @@ public @interface Serial {
    * {@link Version} annotation is missing, serialized form will have serial version 0L assigned.
    */
   @Target({ElementType.TYPE, ElementType.PACKAGE, ElementType.ANNOTATION_TYPE})
+  @Retention(RetentionPolicy.SOURCE)
   public @interface Structural {}
 }


elucash:
Thank you for the PR! I've long hesitated as it might be break something for someone. At the same time CLASS retention to SOURCE change is a lot less riskier that, say, from RUNTIME to either CLASS or SOURCE retained. I'm going with the change for (soon to be released) 2.11. Sorry, it took some time.

jbergfi:
@elucash Thank you for continuing to maintain this library, we've been using it for many years already! We're about to update to the latest but we noticed it did not work quite as before and narrowed it down to this and indeed it broke our setup slightly.

We have a multi-project setup where we have defined the immutables styling/configuration in a core-library that all other projects depend on. This is configured in a package-info.java like this:

```java
@MyImmutableStyle
@Serial.Structural
package com.mybigproject;

import org.immutables.serial.Serial;
```

While we understand that Serial.Structural is really intended to be just for code generation and not visible to the outside it has been convenient to have it only defined in one place rather than each of our projects needing that setup as a boiler-plate. Is there any other way around it other than to copy the package-info.java to each project?

elucash:
Thank you for bringing that up. I'll try to figure something out. My apologies for not being able to calculate all moves like this meta-annotation aspect, which I've overlooked. First thing that comes to mind (except that I don't want to switch retention back and forth continuously) is to, maybe, introduce class-retained `@Serial.AllStructural` or something like that to replace the usage in meta-annotation position.

Issue #1531, opened 2025-01-15
Type annotation handling and TypeMirror#toString


cushon:
JDK 22 and newer fix a bug where type annotations were not being loaded from class files. The bug fix causes `TypeMirror#toString` to correctly include type annotation in some cases where it previously did not.

More information:

* https://www.oracle.com/java/technologies/javase/22-relnote-issues.html#JDK-8225377
* https://bugs.openjdk.org/browse/JDK-8225377

This affects immutables here, because if a type variable has type annotations they will now be included in the string representation:

https://github.com/immutables/immutables/blob/4c737f7fcc3e00799b12e57f3f6abd6b3ebcd800/value-processor/src/org/immutables/value/processor/meta/TypeStringProvider.java#L332

Here's a repro that adapts an existing test, and shows a failure on JDK 22:

```diff
diff --git a/builder/test/org/immutables/builder/fixture/GenericsImprovisedFactories.java b/builder/test/org/immutables/builder/fixture/GenericsImprovisedFactories.java
index bb5a08cf..5a210269 100644
--- a/builder/test/org/immutables/builder/fixture/GenericsImprovisedFactories.java
+++ b/builder/test/org/immutables/builder/fixture/GenericsImprovisedFactories.java
@@ -18,7 +18,7 @@ package org.immutables.builder.fixture;
 import java.util.Set;
 import com.google.common.collect.Iterables;
 import java.util.List;
-import javax.annotation.Nullable;
+import org.checkerframework.checker.nullness.qual.Nullable;
 import org.immutables.builder.Builder;
 import org.immutables.value.Value;

@@ -50,7 +50,7 @@ class GenericsImprovisedFactories {

   @Builder.Factory
   @SuppressWarnings("all")
-  public static <T, V extends RuntimeException> String genericSuperstring(int theory, T reality, @Nullable V evidence)
+  public static <T, @Nullable V extends RuntimeException> String genericSuperstring(int theory, T reality, @Nullable V evidence)
       throws V {
     if (evidence != null) {
       throw evidence;
```

```
$ mvn -version
Apache Maven 3.9.6
Java version: 22.0.2
$ mvn clean package
...
builder/test/org/immutables/builder/fixture/GenericsImprovisedFactories.java:[53,120] It is a compiler/annotation processing bug to receive type variable '@org.checkerframework.checker.nullness.qual.Nullable V' here. To avoid it — do not use not yet generated types in evidence attribute
```

[...]

PR #1532, opened 2025-01-15
Fix a bug with type annotation handling

cushon:
See https://github.com/immutables/immutables/issues/1531


cushon:
Suggestions for a better place to put test coverage for this are welcome.

sdeleuze:
+1 from the Spring team for this fix.

elucash (approved):

elucash:
Thank you for the PR! Got it, I guess the test might be enough (i.e. just not break in the presence of a type annotation)

Issue #1488, opened 2023-10-18
Static factory parameter ordering changed in 2.8.x with inherited parameters


marquiswang:
We noticed a behavioral change in the generated static factory in 2.8.x, when inheriting fields from a parent class.

For example, with the following code:
```
public interface Foo {
    @Value.Parameter
    int a();
}

@Value.Immutable
public interface Bar extends Foo {
    @Value.Parameter
    int b();
}
```

In 2.7.4 and below, Immutables generates 
```
public static ImmutableBar of(int b, int a) {
    return new ImmutableBar(b, a);
}
```

and in 2.7.5 and above, it generates: 
```
public static ImmutableBar of(int a, int b) {
    return new ImmutableBar(a, b);
}
```

I think that the new behavior is better, and in a few cases in our codebase we'd explicitly changed the subclass to be something like the following in order to force the parameter order.

```
@Value.Immutable
public interface Bar extends Foo {
    @Override
    @Value.Parameter
    int a();

    @Value.Parameter
    int b();
}
```

However, we have great concerns about the safety of this migration - if the parameters are all of the same type, then reordering parameters can lead to really dangerous bugs.

Do you have any suggestions for how to migrate to 2.8.x and higher? I'm thinking that some kind of flag that we can pass to Immutables to detect if all fields are the same type AND there is a parent class that provides some fields, then emit a warning unless some annotation is added to the class indicating that it was checked?

elucash:
I was hoping we have such a change under a feature flag. But I cannot find such thing. I might need to look into how it was before and why it's changed (probably when we've discontinued some hacks for older compilers, this might have been a collateral). I hope we can return that older behavior under a feature flag. Will be looking into it, but no estimates :-(

marquiswang:
Thanks @elucash!

I think it was this commit: https://github.com/immutables/immutables/commit/11cc457c7a7400b403f4b8bcebed49dc65e5fa9a

... if that helps at all.

MendelMonteiro:
Is any progress on adding an option to use the older behaviour?

MendelMonteiro:
@elucash  Is this something you would accept a PR for?

elucash:
[...]

Issue #1540, opened 2025-03-29
Generic list pattern with builtinContainerAttributes disabled and deepImmutablesDetection enabled leads to compilation failure in 2.8.0


christophercolumbusdog:
Consider the following classes:

```java
import org.immutables.value.Value;

@Value.Immutable
@Value.Style(
  deepImmutablesDetection = true,
  builtinContainerAttributes = false)
public interface ChildType extends SuperType<ChildType> {
}
```
and
```java
import java.util.List;

public interface SuperType<T extends SuperType<T>> {
  List<T> getNestedChildrenTypes();
}
```

On Immutables version <=2.7.5, it will produce the following for `ImmutableChildType`:
```java
...
public final class ImmutableChildType implements ChildType {
  private final List<ChildType> nestedChildrenTypes;
...
```

However, on Immutables version >=2.8.0, we see the following code that is incorrect and won't compile properly:
```java
public final class ImmutableChildType implements ChildType {
  private final ImmutableChildType nestedChildrenTypes;
```

The `nestedChildrenTypes` is no longer a `List`, and thus will fail compilation.

This failure seems to require both `deepImmutablesDetection = true` and `builtinContainerAttributes = false`. Removing either value from our `@Value.Style` will result in compilation succeeding.

christophercolumbusdog:
Binary searching through releases, I've narrowed down the introduction of this issue to #943  (https://github.com/immutables/immutables/pull/943 / https://github.com/immutables/immutables/commit/5bf91b50472ebc07cf6880d29af63224b86c1f30)
Hopefully this helps in finding a solution/understanding the issue.

christophercolumbusdog:
Restoring the implementation to guarding the `initAttributeValueType` call addresses the issue in my case:
```java
if (supportBuiltinContainerTypes()) {
  initAttributeValueType();
  initImmutableCopyOf();
}
```
(`ValueAttribute`, line 1246)

But I don't have the full context of why it was moved up in the first place. Perhaps @asereda-gs or @elucash could weigh in here.

elucash:
[...]

PR #1541, opened 2025-03-31
Add legacy SourceOrdering option

christophercolumbusdog:
Prior to Immutables 2.7.5, the source ordering was defined be a "shallow" approach where accessors from the child class were added first, then the superclass, then the interfaces. This changed in 2.7.5, where the order became interfaces, then superclass, then child class. Issue 889 acknowledged this as a breaking
change, and it has made it difficult for some clients to migrate safely to later versions.

Add a new system property,
org.immutables.useLegacyAccessorOrdering,
that when set to true, reverts to the <=2.7.4 ordering. This should enable more clients to migrate to more recent versions safely.

This should fix issue #1488

### DIFF
--- a/generator/src/org/immutables/generator/SourceOrdering.java
+++ b/generator/src/org/immutables/generator/SourceOrdering.java
@@ -17,7 +17,6 @@ package org.immutables.generator;
 
 import com.google.common.base.Function;
 import com.google.common.base.Functions;
-import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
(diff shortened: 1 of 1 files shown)


elucash:
Thank you for the PR! It was extremely hard for me to look into it myself and got it right. I assume you've verified it works for the intended use case (migrating from older versions). I want to add at least some tests and planning to add `@Style` attribute in addition to JVM property, so both will work

christophercolumbusdog:
Excellent, thanks for merging! This was tested against our migration use case, and we verified that it enabled the migration in most cases. We ran across issue #1540, which prevented a complete upgrade; hopefully, we can address that one, too!

PR #1543, opened 2025-05-02
Update Generated.java to be documented.

ascopes:
Update the generated annotation to appear in the JavaDoc headers for generated code. This is currently not the case, but is useful information to include to show users of a library using immutables that the type is generated rather than explicitly defined in the project source code.


elucash:
Thank you! merging

PR #1568, opened 2025-07-13
Final stage builder methods' return value should be ignorable

benhalasi:
This change-set should annotate final stage builder interfaces (`BuildFinal`) with [CanIgnoreReturnValue](https://errorprone.info/api/latest/com/google/errorprone/annotations/CanIgnoreReturnValue.html), and allow downstream projects using Error Prone to use Staged Builders without method chaining or disabling the [CheckReturnValue](https://errorprone.info/bugpattern/CheckReturnValue) bug-checker. This aligns with the existing implementation and "normal" and strict builders.

---

Aside from this change, I have a broader question:

We encountered this while trying to propagate optional values to default primitive members in our immutable models. For example, the following model `Foo` will produce a single builder method for `Foo#bar`, `BuildFinal bar(int bar)`, which cannot be called with an `Optional<Integer>` with method chaining, only conditionally with the optional unwrapped.

```
interface Foo {
  default int bar() {
    return 99;
  }
}
```

Have you considered generating "optional" builder methods for default non-optional members? In this example, `BuildFinal bar(Optional<Integer> bar)`?

I've managed to implement this for object types, but not for primitives. I didn't yet find a way to map primitives to their boxed optional types.

If this feature would be welcome, I may give it a go, and ask for you input, pointers on it.

Thanks all!

### DIFF
--- a/value-processor/src/org/immutables/value/processor/Immutables.generator
+++ b/value-processor/src/org/immutables/value/processor/Immutables.generator
@@ -5115,7 +5115,7 @@ private static final long serialVersionUID = [literal serialVersion];
 [/for]
 [/template]
 
-[-- Usefull for annotations etc --]
+[-- Useful for annotations etc --]
 [template eachLine Object... lines][for l in lines]
 [l]
 [/for][/template]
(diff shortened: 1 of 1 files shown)


elucash:
Thank you! Merging

Issue #1572, opened 2025-07-18
Builder copying approach in >= 2.9.0 can lead to performance issues due to method length


christophercolumbusdog:
There is a change in Immutables 2.9.0 that can lead to catastrophic performance regressions for large Immutable classes.

Consider the following class structure:

```java
@Value.Immutable
public interface ParentOne {
    int getAge();

    double getWeight();

    String getName();
}

@Value.Immutable
public interface ParentTwo {
    String getPreferences();

    float getIncome();
}

@Value.Immutable
public interface Child extends ParentOne, ParentTwo {}
```

In Immutables 2.8.8, this resulted in the following `from(..)` method being generated in the buider:
```
private void from(Object object) {
  if (object instanceof ParentOne) {
    ParentOne instance = (ParentOne) object;
    name(instance.getName());
    weight(instance.getWeight());
    age(instance.getAge());
  }
  if (object instanceof ParentTwo) {
    ParentTwo instance = (ParentTwo) object;
    income(instance.getIncome());
    preferences(instance.getPreferences());
  }
}
```

Starting in Immutables 2.9.0, however, the method starts to look much different:
```
private void from(Object object) {
  @Var long bits = 0;
  if (object instanceof Child) {
    Child instance = (Child) object;
    if ((bits & 0x1L) == 0) {
      income(instance.getIncome());
      bits |= 0x1L;
    }
    if ((bits & 0x4L) == 0) {
      name(instance.getName());
      bits |= 0x4L;
    }
    if ((bits & 0x8L) == 0) {
      weight(instance.getWeight());
      bits |= 0x8L;
    }
    if ((bits & 0x2L) == 0) {
      preferences(instance.getPreferences());
      bits |= 0x2L;
    }
    if ((bits & 0x10L) == 0) {
      age(instance.getAge());
      bits |= 0x10L;
    }
  }
  if (object instanceof ParentOne) {
    ParentOne instance = (ParentOne) object;
    if ((bits & 0x4L) == 0) {
      name(instance.getName());
      bits |= 0x4L;
    }
    if ((bits & 0x8L) == 0) {
      weight(instance.getWeight());
      bits |= 0x8L;
    }
    if ((bits & 0x10L) == 0) {
      age(instance.getAge());
      bits |= 0x10L;
    }
  }
  if (object instanceof ParentTwo) {
    ParentTwo instance = (ParentTwo) object;
    if ((bits & 0x1L) == 0) {
      income(instance.getIncome());
[...]

PR #1573, opened 2025-07-21
Optimize mergeInternal generation with specialized methods per subtype

christophercolumbusdog:
This hopes to address #1572 by shortening the `mergeInternal` method by using methods, while maintaining the performance of the bit masking operators, especially for smaller data models.

Prior to this commit, our mergeInternal function that is created when a derived class is used as an Immutables target would check the possible types in a series of if blocks. However, this can lead to an explosion in the length of this method, especially for target classes with multiple parents. Long methods are concerning because it can exceed the JVM's default 8 KB limit for method compilation, forcing calls to remain interpreted.

Instead, generate specialized methods for each relevant subtype. This helps break up the monolithic mergeInternal method while not introducing too many method hops.

This approach, however, requires special handling of our bit masks, since we may have more than one long value to represent cases where the model has >64 fields. To deal with this, introduce a specialized case where we use a long[] of values, in order for each specialized method to easily update the overall state. This approach balances optimizing the simple case (<= 64 fields) for performance on small models, while allowing for larger models with only introducing a minimal amount of overhead in the allocation and access of the array values.

### DIFF
--- a/value-fixture/src/nonimmutables/huge/ChildOfHugeParents.java
+++ b/value-fixture/src/nonimmutables/huge/ChildOfHugeParents.java
@@ -1,0 +1,7 @@
+package nonimmutables.huge;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface ChildOfHugeParents extends HugeParentOne, HugeParentTwo {
+}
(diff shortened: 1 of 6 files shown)


elucash:
Thank you for the PR! Initially I thought of only handling this via static overload approach. But since you've taken some efforts to implement this in the dynamic subtype handling, I think this is good too. I would agree that for attributes count > 64 case, the allocation of small array of longs would not bring noticeable overhead. I'll make some adjustments after merge, for example there's no need to return the array. i.e. it can be done for convenience, but in this case it's no better than just void methods.

elucash:
after checkout out code it, I see there's no problem with return, I've just confused something

PR #1574, opened 2025-07-21
fix: set correct override for Jackson annotation processor option

devinrsmith:


elucash:
I believe it is. Thank you for PR! merging

PR #1596, opened 2025-09-19
Add test for Default.{Int, Float, ...}

fp7:


elucash:
incorporated tests like these into the codebase

Issue #1600, opened 2025-09-26
Immutable 2.11.4 causes StackOverflowError when object uses Jackson and Guava GwtCompatible annotations


pkoenig10:
The following example compiles successful when using Immutables 2.11.3 but fails with `StackOverflowError` when using Immutables 2.11.4.

```java
import com.fasterxml.jackson.annotation.JsonProperty;
import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
import com.fasterxml.jackson.databind.annotation.JsonSerialize;
import com.google.common.annotations.Beta;
import org.immutables.value.Value;

@Value.Immutable
@JsonSerialize(as = ImmutableMyValue.class)
@JsonDeserialize(as = ImmutableMyValue.class)
public interface MyValue {

    @Beta
    @JsonProperty("foo")
    String foo();
}
```

```
Caused by: java.lang.StackOverflowError
        at jdk.compiler/com.sun.tools.javac.code.Symbol$ClassSymbol.getRawAttributes(Symbol.java:1379)
        at jdk.compiler/com.sun.tools.javac.code.Symbol.getAnnotationMirrors(Symbol.java:734)
        at jdk.compiler/com.sun.tools.javac.code.Symbol$TypeSymbol.getAnnotationMirrors(Symbol.java:798)
        at org.immutables.value.internal.$processor$.meta.$Proto.isAnnotatedWith($Proto.java:2207)
        at org.immutables.value.internal.$processor$.meta.$Annotations.annotationTypeMatches($Annotations.java:130)
        at org.immutables.value.internal.$processor$.meta.$Annotations.annotationTypeMatches($Annotations.java:209)
        at org.immutables.value.internal.$processor$.meta.$Annotations.annotationTypeMatches($Annotations.java:209)
        at org.immutables.value.internal.$processor$.meta.$Annotations.annotationTypeMatches($Annotations.java:209)
        at org.immutables.value.internal.$processor$.meta.$Annotations.annotationTypeMatches($Annotations.java:209)
        ...
```

This appears to be a result of the changes in https://github.com/immutables/immutables/commit/3da32c580d3526584bd0b909cf891639b2cf214f, which doesn't seem to correctly handle  annotations that are applied to themselves, like [`@GwtCompatible`](https://github.com/google/guava/blob/8868c096cfdabbe38170b6e395369c315cfb72a1/guava/src/com/google/common/annotations/GwtCompatible.java#L35-L36) (which is applied to `@Beta` and other Guava annotations).

PR #1601, opened 2025-10-02
#1600 Stop infinite recursion for annotations applying to themselves

aldexis:
This fixes https://github.com/immutables/immutables/issues/1600 which was a bug introduced by https://github.com/immutables/immutables/commit/3da32c580d3526584bd0b909cf891639b2cf214f.

I'm honestly not too sure about where to put the test to reproduce the bug and validate it. Ultimately all that matters is that we test the compilation of the class, but I couldn't figure out where other examples of tests like this would live, so I pattern-matched from the test added in that same commit.

Happy to update the test to whatever makes sense!

### DIFF
--- a/value-fixture/src/org/immutables/fixture/annotation/ExampleModelWithSelfApplyingAnnotation.java
+++ b/value-fixture/src/org/immutables/fixture/annotation/ExampleModelWithSelfApplyingAnnotation.java
@@ -1,0 +1,14 @@
+package org.immutables.fixture.annotation;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import nonimmutables.SelfApplyingAnnotation;
+import org.immutables.value.Value;
+
+@Value.Immutable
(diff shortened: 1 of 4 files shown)


value-fixture/src/org/immutables/fixture/annotation/ExampleModelWithSelfApplyingAnnotation.java
```diff
@@ -0,0 +1,15 @@
+package org.immutables.fixture.annotation;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import nonimmutables.SelfApplyingAnnotation;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@Value.Style(passAnnotations = {SelfApplyingAnnotation.class})
```
aldexis, line 9: I had to add this for the test to pass (I somehow thought I had it passing before - not sure what happened). It doesn't really matter though, since what we want to test is the fact that it compiles fine, though I'm not sure how to best set this up

aldexis:
Thanks for merging!

PR #1606, opened 2025-10-15
Fix `.unset*()` for modifiable with primitive field and default value.

aquariusrick:
### Summary
Ensure modifiable with primitive field and default value returns the default value 
and `.*IsSet()` returns `false` after `.unset*()` is called.

### Description:
Currently a field such as:
```java
@Value.Modifiable
interface Standalone {

  @Value.Default
  default int def() {
    return 1;
  }
}
```

Would generate the following for the `.unset*()` method:
e.g.
```java
/**
 * Reset an attribute to its initial value.
 * @return {@code this} for use in a chained invocation
 */
@CanIgnoreReturnValue
public final ModifiableStandalone unsetDef() {
  optBits |= 0;
  def = 0;
  return this;
}
```
Having `optBits |= 0;` does nothing, and doesn't effectively unset the field.  Thus, subsequent calls to `.def()` would return `0`, instead of the value returned by the method annotated with `@Value.Default` (i.e. `1`).  Further, subsequent calls to `.defIsSet()` would return `true` when it should return `false`.

This change would instead generate the following for the `.unset*()` method:
e.g.

```java
/**
 * Reset an attribute to its initial value.
 * @return {@code this} for use in a chained invocation
 */
@CanIgnoreReturnValue
public final ModifiableStandalone unsetDef() {
  optBits &= ~OPT_BIT_DEF;
  def = 0;
  return this;
}
```

Having `optBits &= ~OPT_BIT_DEF` will cause subsequent calls to `.defIsSet()` to return `false`, and calls to `.def()` to invoke the method annotated with `@Value.Default` to get the value.

### DIFF
--- a/value-fixture/src/org/immutables/fixture/modifiable/Companion.java
+++ b/value-fixture/src/org/immutables/fixture/modifiable/Companion.java
@@ -111,6 +111,11 @@ public interface Companion {
     }
 
     @Value.Default
+    default int def2() {
+      return 0;
+    }
+
+    @Value.Default
     default String defs() {
       return "";
     }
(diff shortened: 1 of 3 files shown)


elucash:
Thank you for PR! this is nice finding, definitely has to be fixed. Merging

Issue #1609, opened 2025-10-21
Jackson 3 compatibility


mikebell90:
1. Jackson 3 changes packages and gav to allow side-by-side
2. While most annotations remain in the Jackson 2 package, to allow compatibility the databind ones, such as `JsonSerialize` or `JsonDeserialize` do not
3. This breaks Immutables

A colleague of mine at our company identified and has a tentative PR. However we didn't want to allocate time to finish integrating with test framework (which he described as extensive ;) @elucash ) until we knew

* Known issue
* PR welcome

mcliedtke:
Jackson 3 release notes ref: https://github.com/FasterXML/jackson/wiki/Jackson-Release-3.0

The main take away and what impacts immutables is mainly the new java package spaces Jackson 3 uses. Jackson-annotations keeps the namespace `com.fasterxml.jackson` but databind and core move to `tools.jackson`. The majority of Jackson annotations are defined in Jackson-annotations so remain untouched, but there are two annotations that are part of databind that moved from `com.fasterxml.jackson` to `tools.jackson`.
* `com.fasterxml.jackson.databind.annotation.JsonDeserialize` -> `tools.jackson.databind.annotation.JsonDeserialize`
* `com.fasterxml.jackson.databind.annotation.JsonSerialize` ->  `tools.jackson.databind.annotation.JsonSerialize`

Immutables seems to have logic to auto-detect presence of @JsonDeserialize/@JsonSerialize (i.e. "To enable this, you should use @JsonSerialize or @JsonDeserialize annotation." https://immutables.github.io/json.html). Immutables needs to be updated to recognize `@JsonSerialize` or `@JsonDeserialize` from either of the `com.fasterxml.jackson` or `tools.jackson` package spaces, basically just treating them as aliases of one another for triggering the logic for "jackson annotation detection".

PR #1615, opened 2025-11-05
Jackson 3 support for Immutables

dkaukov:
Fixes: https://github.com/immutables/immutables/issues/1609

- Make generated Immutables work with Jackson 3.
- Still works with Jackson 2.x: the generator mirrors whichever Jackson annotations you use on the interface (2.x or 3.x), so both should work in theory.
- Tests: none yet. I need pointers on where to add a test fixture for this (which module/folder and any existing examples I should follow).


elucash:
Indeed, I've checked the annotations, I was not sure those are coming from `jackson-databind` (are actually from `jackson-annotations`, right). Thank you for reminding!

The only piece in question I can find is the following template

```
[template overrideJsonDeserialize Type type][if type.jacksonDeserialized]@[jackson].databind.annotation.JsonDeserialize[/if][/template]
```

I cannot clearly recollect what it does, maybe something trivial, I'll get back to this when I'll have a chance

mikebell90:
To be clear, do you have immediate requirements for Dmitry? I can check with him to make sure what you commented on tests is sufficient and ask him to post accordingly. Or should we pause awaiting your checking into the templating.

elucash:
```
bad class file: /home/runner/.m2/repository/tools/jackson/core/jackson-databind/3.0.1/jackson-databind-3.0.1.jar(/tools/jackson/databind/annotation/JsonDeserialize.class)
    class file has wrong version 61.0, should be 55.0
```

@dkaukov @mikebell90 If Jackson3 raises compatibility to JDK 17 (class file version 61 (?)), then we need to put new `org.immutables.fixture.jackson3` packages into corresponding [`src-java-17`](https://github.com/immutables/immutables/tree/master/value-fixture/src-java-17) and [`test-java-17`](https://github.com/immutables/immutables/tree/master/value-fixture/test-java-17) folders

elucash:
Thank you! Everything passing. I'll merge it as as soon as I ready (tonight or tomorrow), it will get to the published release (hopefully) this weekend

PR #1627, opened 2025-12-13
JDK 23+.

SimY4:
Enable immutable compilation on JDK 23+

All the learnings from this work:

- Since Java 23 unless specified explicitly in `<annotationProcessor>` configuration section annotation processors won't be located on classpath and executed without providing `-proc:full` compiler flag. This change was made in JDK 23.

- error prone can't compile JDK 25 target without an upgrade. But upgrade brings backward incompatible changes to error prone configuration. So I've disable error prone for JDK targets beyond JDK 23. Once immutables will bump minimum supported JDK version, error prone can be updated and enabled again for JDKs past 23.

- ecj compiler is not running annotation processors. I recon the issue is similar to `-proc:full` flag in my point 1 in findings, except I have no clue how to provide this to `ecj`. I tried to look up any info on that, no luck. So I've excluded `ecj` from build matrix until I can figure out how to bring it back.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,9 +14,12 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ '8', '11', '17', '21'  ]
+        java: [ '8', '11', '17', '21', '25'  ]
         compiler: [ 'javac', 'ecj' ]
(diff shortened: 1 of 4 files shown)


elucash:
Thank you! Seems only new, good stuff added!

PR #1644, opened 2026-03-13
#1642 Fix for further problems to find nested elements on plain element with nested elements with mongo backend

### LINKED ISSUE #1642: Cannot combine criteria inside .any().with(any.<plain-element>.innerany.any()…) for MongoDB
Based on [bug 1623](https://github.com/immutables/immutables/issues/1623) we have found another edge case.
Immutables Criteria for MongoDB seems to have a bug, when you try to find a document with an element inside a nested collection on a plain element with another nested collection that matches a condition. For example:
```
person.pets.any()
  .with(pet -> pet.collar.tags.any().is("rescue"))
```
Executing such a criterion against an in-memory backend works.
However, running such a criterion against a mongo backend a wrong bson filter is created and so the results are wrong.

The exptected bson filter should be:
`Filter{fieldName='pets.collar.tags', value=rescue}`
But the generated bson filter is:
`Filter{fieldName='pets.tags', value=rescue}`

I'm not the way to create a test to reproduce the error and to fix it.

rgatbert:
Demonstrates and fixes #1642

### DIFF
--- a/criteria/mongo/test/org/immutables/criteria/mongo/FindVisitorTest.java
+++ b/criteria/mongo/test/org/immutables/criteria/mongo/FindVisitorTest.java
@@ -18,6 +18,7 @@ package org.immutables.criteria.mongo;
 
 import org.immutables.criteria.Criterias;
 import org.immutables.criteria.matcher.PresentAbsentMatcher;
+import org.immutables.criteria.personmodel.PersonCriteria;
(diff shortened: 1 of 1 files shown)


asereda-gs:
Thanks for this PR.

Can you please add tests in `FindVisitorTest` to confirm expression tree?

PR #1658, opened 2026-05-13
#1655 Fix TYPE_USE @Nullable on concrete array types

### LINKED ISSUE #1655: TYPE_USE @Nullable on concrete array types produces warnings
Does #1579 only cover TYPE_USE array annotations for generic type variables, and not concrete types like `byte` and `String`?

Repro:

```java
public interface NullableArrays {
  byte @Nullable [] primitiveArray();
  String @Nullable [] referenceArray();

  @Value.Immutable
  @Value.Style(jdkOnly = true)
  interface Child extends NullableArrays {
    @Override byte @Nullable [] primitiveArray();
    @Override String @Nullable [] referenceArray();
  }
}
```

`mvn clean compile -pl value-fixture -am` on Java 21 produces:
```
warning: (immutables) Type parsing problem in FromSupertypesModel:
  Cannot parse type from input string 'byte @org.jspecify.annotations.Nullable []'.
  unexpected trailing terms '@org.jspecify.annotations.Nullable[]'

warning: (immutables:from) Generated builder '.from' method will not copy
  from attribute 'primitiveArray' / 'referenceArray'
```

`byte @Nullable []` gets both the parsing error and `.from` warning. `String @Nullable []` gets only the `.from` warning.

Immutables 2.12.2-SNAPSHOT (repo HEAD), JSpecify 1.0.0, Java 21.

viditochani:
Fixes #1655

`byte @Nullable []` and `String @Nullable []` produced warnings because:

1. The type parser in `FromSupertypesModel` did not handle `@` annotations between the element type and array brackets
2. The source access workaround in `TypeStringProvider` checked for `@Nullable[]` (no space) but the valid Java syntax is `@Nullable []` (with space)

Generic type variables (`V @Nullable []`) already worked; this fixes concrete types.

### Changes
- `Type.java`: consume `@` annotations before `[]` in the array parsing loop
- `TypeStringProvider.java`: also match `@Nullable []` with a space
- Added `NullableArrays.java` test fixture with `ChildOverrides` (parent declares arrays, child overrides) and `ChildIntroduces` (child adds arrays not in parent)
- Added test in `TypeuseAnnotationTest` verifying `.from()` copies nullable arrays from supertype and null arrays are accepted

### Testing
`mvn clean test -pl value-fixture -am` — all 323 tests pass, no new warnings.

### DIFF
--- a/value-fixture/src/org/immutables/fixture/nullable/typeuse/NullableArrays.java
+++ b/value-fixture/src/org/immutables/fixture/nullable/typeuse/NullableArrays.java
@@ -1,0 +1,41 @@
+package org.immutables.fixture.nullable.typeuse;
+
+import org.immutables.value.Value;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Reproducer for TYPE_USE nullable annotations on arrays.
+ *
+ * Case 1: Parent declares the array methods, child overrides.
+ * Case 2: Parent does NOT declare array methods, child adds them.
+ */
+public interface NullableArrays {
+
+  // Parent has these
+  byte @Nullable [] primitiveArray();
+  String @Nullable [] referenceArray();
+
+  // Case 1: child overrides parent's array methods
+  @Value.Immutable
+  @Value.Style(jdkOnly = true)
+  interface ChildOverrides extends NullableArrays {
+    @Override
+    byte @Nullable [] primitiveArray();
+
+    @Override
+    String @Nullable [] referenceArray();
+  }
+
+  // A parent without array methods
+  interface EmptyParent {
+    String name();
+  }
+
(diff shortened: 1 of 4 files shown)


elucash:
Thank you for the PR! I think changes are overall spot on! For some reason, this test is failing

``` java
NullableArrays.ChildOverrides withNulls = ImmutableChildOverrides.builder().build()
```

elucash:
Ok, it seems I've found a workaround, the problem is only for ECJ compiler (`-Pecj`). Merging this and will apply a fix on top

viditochani:
Thanks for the quick merge! I only tested with javac, so I missed the ECJ issue. Thanks for catching and fixing that.

Issue #1663, opened 2026-05-21
Unexpected behaviour for Derived field depending on Default Optional fields


jbergfi:
Thanks again for maintaining this great library! We've just struck into something that caused a bug for us, consider the following test:

````java
import java.util.Optional;

import org.immutables.value.Value;
import org.immutables.value.Value.Immutable;
import org.junit.Assert;
import org.junit.Test;

public class ImmutableTest {

	@Immutable
	static abstract class AbstractDerivedTest {
		abstract int foo();
		@Value.Derived int derived() { return foo() - bar().orElse(0); }
		@Value.Default Optional<Integer> bar() { return Optional.of(5); }
	}
	
	@Test
	public void testDerived() throws Exception {
		Assert.assertEquals(-1, DerivedTest.of(1, Optional.of(2)).derived());
	}
}

````

The documentation says Default and Derived can depend on each other so this was expected to work and the derived should be able to utilize both foo and bar as set in constructor. Surprisingly it does not pick up bar value (or sees it as Optional.empty()) and test fails with 1 instead of succeeding with -1. I've come to the conclusion that it's becuase the default is an Optional too. If I either remove @Value.Default or Optional the example works (but loosing some functionality). It also work if I re-order and put the derived last in the definition.

I recognise it's a bit special edge case to mix both @Value.Default and Optional together this way but it took quite a while to understand what was going on when having this in production code. We'll probably workaround the issue but you might want to fix this in a future release.

Tested with Immutables 2.12.1

**Comment by elucash** [comment]
Implementation of the generate getter seems to be wrong, it doesn't account for `isDefault`, only `isOptional`

```java
  /**
   * @return The computed-at-construction value of the {@code derived} attribute
   */
  @Override
  int derived() {
    InitShim shim = this.initShim;
    return shim != null
        ? shim.derived()
        : this.derived;
  }

  /**
   * @return The value of the {@code bar} attribute
   */
  @Override
  Optional<Integer> bar() {
    return Optional.ofNullable(bar);    // <-- this should go through this.initShim as above
  }
```

elucash:
I've added wip commit, just to have a context later, but need to add some tests etc. If you can help with that it would great.

jbergfi:
@elucash I added a rudimentary test in a PR that tests this. It also seems your initShim fix did the trick so the test now succeeds.

elucash:
@jbergfi Thank you for PR! The use case works correctly now

PR #1664, opened 2026-05-22
Test for #1663, derived attribute depending default optional attribute

jbergfi:
Test for https://github.com/immutables/immutables/issues/1663. Fails with Immutables 2.12.2, succeeds with Immutables 2.12.3-SNAPSHOT


elucash:
Thank you for PR!

PR #1669, opened 2026-07-01
#1668 ECJ: Skip Javadoc section when searching for `extends` keyword

### LINKED ISSUE #1668: ECJ only: Generated `Builder` class is marked `final` during codegen despite having a parent builder class defined in source code
When compiling with ECJ (i.e. the Eclipse compiler via the VS Code Red Hat Java plugin's JDTLS), Immutables doesn't correctly detect the parent-of-builder pattern, and the generated `ImmutableXYZ.Builder` class ends up `final`, despite the `XYZ.Builder` class extending it.

Example of problematic code using parent-of-builder pattern:

```
package test;

/**
 * Inputs.
 */
@Value.Immutable
@Style(visibility = ImplementationVisibility.PACKAGE, overshadowImplementation = true)
public abstract class Inputs {

    /**
     * Data
     *
     * @return data
     */
    public abstract String data();

    /**
     * Create a new {@link Inputs.Builder}.
     *
     * @return new builder
     */
    public static Builder builder() {
        return new Builder();
    }

    /**
     * Builder for {@link Inputs}.
     */
    public static class Builder extends ImmutableInputs.Builder { }

}
```

While the codegen does run successfully in JDTLS (which is great!), I end up with a compilation error in my source code, like:

```
The type Builder cannot subclass the final class ImmutableInputs.Builder - Java(16777529)
```

Inspecting the generated `Builder` class, it's marked `final` where it's not supposed to be:

```
  @Generated(from = "Inputs", generator = "Immutables")
  @NotThreadSafe
  public static final class Builder extends Inputs.Builder {
```

The weird thing was, this used to work for me. When my codebase had less code, Immutables was working as expected. At some point, as the codebase grew, it stopped working.

I've managed to narrow down the problem. I found that there are some ECJ- (Eclipse) specific tricks that Immutables uses to get all the required info during annotation processing (required because of ECJ's multi-phase compilation - not all source info is available in all phases).

One of the tricks involves parsing the parent builder's source code, searching for the `extends` declaration. If the `extends` keyword is there, the generated `Builder` shouldn't be marked `final`. The way it's implemented is by searching from the start of the declaration until it hits a `{`, indicating the class declaration is over.

But the way it's currently implemented, the declaration also includes any Javadoc that comes before the class declaration starts! **So if the Javadoc just so happens to contain a `{` (which mine frequently does), the Eclipse-specific code doesn't work - it never finds the `extends` keyword, and the `Builder` is always marked `final`, which will cause a compile error.**

In summary - it's an ECJ-specific problem, and it only comes up when your parent-of-builder class has Javadoc that happens to contain a `{` symbol.

**Immutables version used**: 2.10.0 (same problem with 2.12.2)

I can open a PR to try and fix this.

tyner-ibm:
Fixes https://github.com/immutables/immutables/issues/1668

### Problem

A typical parent-of-builder class declaration looks like:
```
static class Builder extends ImmutableParentOfBuilder.Builder { }
```

That means the generated `ImmutableParentOfBuilder.Builder` class can't be marked `final` or compilation will fail. 

In ECJ mode, we're searching for the `extends` keyword to decide if the generated `Builder` shouldn't be marked `final`. That normally works fine.

However, when the declaration looks like this, we were accidentally including the Javadoc as part of the search string, and stopping the search at the first `{`. 
```
    /**
     * Builder class for {@link ParentOfBuilder}.
     */
    static class Builder extends ImmutableParentOfBuilder.Builder { }
```

In this case, the `extends` keyword wouldn't get detected, and the `Builder` would get marked `final`, causing a compile error.

### Solution

This PR changes the approach from "start at the Javadoc, search for `extends`, stop at the first `{`" to "start at the Javadoc, stop at the start of the body, strip out comments and Javadoc, then search for `extends`".

### Testing

Added a unit test - the test fails without the fix (when using `-Pecj`), and the test passes with the fix

Using custom build of `2.12.3-SNAPSHOT` fixes the Eclipse JDTLS issue seen in https://github.com/immutables/immutables/issues/1668

### DIFF
--- a/generator/src/org/immutables/generator/SourceExtraction.java
+++ b/generator/src/org/immutables/generator/SourceExtraction.java
@@ -260,7 +260,7 @@ public final class SourceExtraction {
 
     private static CharSequence extractSuperclass(SourceTypeBinding binding) {
       CharSequence declaration = readSourceDeclaration(binding);
-      StringTokenizer tokenizer = new StringTokenizer(declaration.toString(), "<>, \t\n\r", true);
+      StringTokenizer tokenizer = new StringTokenizer(declaration.toString(), "<>{}, \t\n\r", true);
       int genericsOpened = 0;
       while (tokenizer.hasMoreTokens()) {
         String t = tokenizer.nextToken();
@@ -306,7 +306,7 @@ public final class SourceExtraction {
 
       // Use modifiersSourceStart as the start bound: skip past any Javadoc comment that
       // precedes the declaration.
-      // Use bodyStart as the end bound: it's the exact position of the '{' that opens
+      // Use bodyStart as the end bound: it's the position just after the '{' that opens
       // the class body.
       int start = referenceContext.modifiersSourceStart;
       int end = referenceContext.bodyStart;
--- a/generator/src/org/immutables/generator/SourceExtraction.java
+++ b/generator/src/org/immutables/generator/SourceExtraction.java
@@ -259,7 +259,11 @@ public final class SourceExtraction {
     }
 
     private static CharSequence extractSuperclass(SourceTypeBinding binding) {
-      CharSequence declaration = readSourceDeclaration(binding);
+      // Strip all comments before tokenizing so that '{', '}', or 'extends' inside
+      // or next to a Javadoc or line comment does not affect the scan.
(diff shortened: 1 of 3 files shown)


elucash:
Thank you for the PR! This is a good fix

PR #1670, opened 2026-07-02
Fix #1665: prevent synthetic nullability from leaking onto mandatory attributes

### LINKED ISSUE #1665: Nullable erroneously being attached to getters on immutable classes
Good afternoon!

Spotted a slightly odd problem in how `@Nullable` is being applied to generated code.

I am using the following configuration:

```java
@NullMarked
@Style(
    defaults = @Immutable(copy = false),
    defaultAsDefault = true,
    get = {"get*", "is*"},
    jdkOnly = true,
    jdk9Collections = true,
    nullableAnnotation = "org.jspecify.annotations.Nullable",
    passAnnotations = Nullable.class,
    validationMethod = ValidationMethod.MANDATORY_ONLY
)
package myapp;

import org.immutables.value.Value.Immutable;
import org.immutables.value.Value.Style;
import org.immutables.value.Value.Style.ValidationMethod;
import org.jspecify.annotations.NullMarked;
import org.jspecify.annotations.Nullable;
```
and have the following class defined:
```java
@Value.Immutable
public interface CommandInvocation {
  Command getRootCommand();

  @Nullable
  SubCommand getSubCommand();
}
```

For some reason, on Immutables 2.12.1 and 2.12.2, I am seeing the JSpecify `@Nullable` get attached to the `getRootCommand()` method on the implementation class despite it not being marked as `@Nullable`. Alongside the semantics of the `@NullMarked` package, I'd expect this to only apply `@Nullable` to interface methods that were explicitly flagged as being `@Nullable`.

```java
/**
 * Immutable implementation of {@link CommandInvocation}.
 * <p>
 * Use the builder to create immutable instances:
 * {@code ImmutableCommandInvocation.builder()}.
 */
@Generated(from = "CommandInvocation", generator = "Immutables")
@SuppressWarnings({"all"})
@javax.annotation.processing.Generated("org.immutables.processor.ProxyProcessor")
public final class ImmutableCommandInvocation
    implements CommandInvocation {
  private final @Nullable CommandDescriptor rootCommand;   // Shouldn't be @Nullable!
  private final @Nullable SubCommandDescriptor subCommand;
  
  ...

  /**
   * @return The value of the {@code rootCommand} attribute
   */
  @Override
  public @Nullable CommandDescriptor getRootCommand() {  // Shouldn't be @Nullable!
    return rootCommand;
  }

  /**
   * @return The value of the {@code subCommand} attribute
   */
  @Override
  public @Nullable SubCommandDescriptor getSubCommand() {
    return subCommand;
  }
```

This is a little strange to me as I've been using JSpecify with Immutables elsewhere and have yet to see an issue like this.

My main issue is that my project now fails ErrorProne with Uber's NullAway because it wants me to explicitly check that `.getRootCommand()` returns a non-null value when this is always guaranteed to be the case.

Any help would be greatly appreciated 🙂

KingRayuga:
### What was happening
Under `@NullMarked` (JSpecify) combined with `validationMethod = MANDATORY_ONLY`, mandatory reference attributes were being generated with a typeuse `@Nullable` on both the field and the accessor even though they aren't actually nullable. This tripped up NullAway with false positives.

### Why
Mandatory attributes get an internal "synthetic" nullability that's only meant for builder/validation logic. That synthetic marker was slipping through into the generated code as a real `@Nullable` annotation.

### The fix
Guarded `whenTypeuseFallbackNullable()` with `!isNullabilitySynthetic()`, so only attributes that are *explicitly* marked `@Nullable` emit the type-use annotation. Synthetic-only nullability stays internal, where it belongs.

### Tests
Added a regression test that verifies:
- mandatory attributes are **not** annotated `@Nullable`
- explicitly nullable attributes **still are**

### DIFF
--- a/value-fixture/src/org/immutables/fixture/nullable/typeuse/MandatoryOnlyNullMarked.java
+++ b/value-fixture/src/org/immutables/fixture/nullable/typeuse/MandatoryOnlyNullMarked.java
@@ -1,0 +1,29 @@
+package org.immutables.fixture.nullable.typeuse;
+
+import org.immutables.value.Value;
+import org.immutables.value.Value.Style.ValidationMethod;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Reproduces immutables/immutables#1665: under {@code @NullMarked} (JSpecify) combined with
+ * {@code validationMethod = MANDATORY_ONLY}, mandatory reference attributes get an internal
+ * "synthetic" nullability which must NOT leak onto the generated field/accessor as a type-use
+ * {@code @Nullable}. Only attributes explicitly marked {@code @Nullable} should be nullable.
+ */
+@NullMarked
+@Value.Immutable
+@Value.Style(
+    jdkOnly = true,
+    defaultAsDefault = true,
+    get = {"get*", "is*"},
+    nullableAnnotation = "org.jspecify.annotations.Nullable",
+    validationMethod = ValidationMethod.MANDATORY_ONLY)
+public interface MandatoryOnlyNullMarked {
+  /** Mandatory attribute: must never be annotated {@code @Nullable} in generated code. */
+  String getMandatory();
+
+  /** Explicitly nullable attribute: must be annotated {@code @Nullable} in generated code. */
+  @Nullable
+  String getOptional();
+}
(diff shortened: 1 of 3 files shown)


elucash:
Thank you for the PR! Really good catch!

PR #1671, opened 2026-07-07
Fix #1662: propagate JSpecify NullMarked annotation to record builder stages

### LINKED ISSUE #1662: records with staged builder do not propagate JSpecify non-null metadata to generated BuilderStages interfaces
### Summary
When using `@Value.Builder` on a Java record together with:
- `@Value.Style(stagedBuilder = true)`
- `@NullMarked`
- `org.jspecify.annotations.Nullable`

the generated staged builder interfaces (`*BuilderStages`) appear to lose the non-null-by-default semantics from `@NullMarked`.
As a result, generated stage methods for required attributes accept `null` and IDE / static analysis does not complain, even though the corresponding record components are non-null and the builder enforces them at runtime.
This makes the staged builder less useful for null-safety.

### Version
- Immutables: `2.12.2`
- Java: `21`
- JSpecify: `1.0.0`
### Minimal code to reproduce
```java
package demo;
import org.jspecify.annotations.NullMarked;
import org.jspecify.annotations.Nullable;
@Value.Style(stagedBuilder = true)
@Value.Builder
@NullMarked
public record Person(
    String name,
    @Nullable String nickname,
    int age
) {
  public static PersonBuilderStages.BuildStart builder() {
    return PersonBuilderStages.start();
  }
}



// calling side:

Person.builder()
.name(null) // <- this should be flagged by IDE / static analysis as invalid
.age(42)
.build();
```

The generated staged builder interface for the required attribute looks like this:
```java
public interface NameBuildStage {
  AgeBuildStage name(String name);
}
```

what i expected:
```java
@NullMarked
public interface NameBuildStage {
  AgeBuildStage name(String name);
}
```

or

```java
public interface NameBuildStage {
  AgeBuildStage name(@org.jspecify.annotations.NonNull String name);
}
```

This issue seems to occur specifically for generated staged builder interfaces of java records.  
When using "classic" immutables stage builder with interfaces i could not reproduce that issue.

KingRayuga:
## Changes

- Propagating the JSpecify NullMarked annotation to record builder stages
- Added testcases to verify the behaviour

### DIFF
--- a/value-fixture/src-java-17/org/immutables/fixture/j17/RecordStagedNullMarked.java
+++ b/value-fixture/src-java-17/org/immutables/fixture/j17/RecordStagedNullMarked.java
@@ -1,0 +1,14 @@
+package org.immutables.fixture.j17;
+
+import org.immutables.value.Value;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+@Value.Builder
+@Value.Style(stagedBuilder = true)
+public record RecordStagedNullMarked(String name, @Nullable String nickname, int age) {
+  public static RecordStagedNullMarkedBuilderStages.BuildStart builder() {
+    return RecordStagedNullMarkedBuilderStages.start();
+  }
+}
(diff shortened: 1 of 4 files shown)


elucash:
Thank you for the PR! I've yet to look at it in detail, but I see build failed with:

```
Error:    RecordStagedNullMarkedTest.optionalStageParamIsNullable:19 » NoSuchMethod org....
```

KingRayuga:
Thanks, I noticed it too I am trying to reproduce this in my local to see why the build is failing

KingRayuga:
I have fixed the testcase and I have dropped the attribute level assertion because I think the behaviour is little different with ECJ although the main class level null marked assertion is still there

elucash:
Thank you for the PR! merging!

PR #1673, opened 2026-07-29
Name attribute builder helpers after their own attribute

### LINKED ISSUE #1672: Generated code is not reproducible: attribute-builder helper names leak between attributes through a static cache
When `attributeBuilderDetection` is enabled and one buildable type is reachable under two attribute names, the generated helper takes its name from whichever attribute the processor analysed first. That depends on compilation order, so identical sources produce different generated code between runs.

Affects 2.12.2 and current master (verified identical at 45a9d4c5).

## Reproduction

Plain `javac`, no build tool.

```java
// ReproStyle.java
package repro;

import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

@Retention(RetentionPolicy.CLASS)
@Value.Style(attributeBuilderDetection = true)
public @interface ReproStyle {}
```

```java
// Inner.java, Alpha.java, Beta.java - one type per file, same package
@Value.Immutable
public interface Inner {
    String v();
}

@ReproStyle
@Value.Immutable
public interface Alpha {
    Inner alphaInner();
}

@ReproStyle
@Value.Immutable
public interface Beta {
    Inner betaInner();
}
```

Compile twice, varying only the order the two value types are handed to `javac`:

```
javac -proc:full -cp value-2.12.2.jar -processorpath value-2.12.2.jar -d out1 -s gen1 \
  ReproStyle.java Inner.java Alpha.java Beta.java

javac -proc:full -cp value-2.12.2.jar -processorpath value-2.12.2.jar -d out2 -s gen2 \
  ReproStyle.java Inner.java Beta.java Alpha.java
```

## Result

| Order | helpers in `ImmutableAlpha` | helpers in `ImmutableBeta` |
|---|---|---|
| `Alpha` first | `alphaInner*` | **`alphaInner*`** |
| `Beta` first | **`betaInner*`** | `betaInner*` |

`gen1/repro/ImmutableBeta.java` and `gen2/repro/ImmutableBeta.java` differ, from identical sources. `ImmutableBeta` has no attribute called `alphaInner`, yet in the first ordering its helpers are named after one.

Expected: helpers named consistently, and output that does not depend on compilation order.

## Cause

The helper is emitted once per buildable type, but named after an attribute:

```
[template convertToValueType AttributeBuilder n][n.attributeName]ValueOf[/template]
```

Which attribute supplies the name depends on which is analysed first, because of the cache in `AttributeBuilderReflection`:

```java
private static final Map<String, AttributeBuilderDescriptor> analyzedReturnTypes = new HashMap<>();   // :27

private static String cachingKey(ValueAttribute valueAttribute) {                                     // :31-34
  return valueAttribute.containedTypeElement.getQualifiedName()
      + Arrays.toString(valueAttribute.containingType.constitution.style().attributeBuilder());
}
```

The key omits the attribute name, so the first attribute of a type to be analysed fixes the descriptor - including its `attributeName` - for every later attribute of that type. The map is `static` and never cleared, so its lifetime is the processor's classloader. In a long-lived compiler process such as a Gradle daemon, the winner can be decided by a module compiled earlier.

The mismatch between "one helper per type" and "named after an attribute" is the defect; the cache is what makes it visible.

## Fix

Name the helpers after the value type, so the name cannot depend on analysis order. `AttributeBuilderDescriptor.getQualifiedValueTypeNameWithUnderscores()` already exists for this and is unreferenced. The cache is left alone, so detection behaviour is unchanged. #1674 has the fix and a regression test.

## Why it matters

Generated code becomes non-reproducible: a jar containing the generated classes changes bytes between builds of identical source, so downstream tasks that track it on a classpath miss the build cache. The flip only happens when compilation order changes, so it is invisible on a developer machine and shows up across CI agents, which makes it expensive to diagnose.

pbielicki:
Fixes #1672.

`analyzedReturnTypes` is keyed by the contained type and the `attributeBuilder` discovery pattern, but the cached `AttributeBuilderDescriptor` carries `attributeName`. The first attribute of a given type to be analysed therefore named the attribute-builder helpers for every later attribute of that type, and since the map is `static` the winner depended on compilation order — so identical sources produced different generated code between runs.

Adding the attribute name to the key gives each attribute its own descriptor. It also stops a negative verdict (`analyzedReturnTypes.put(cacheKey, null)`) reached through one attribute from being reused for the others.

The first commit adds a failing test, the second contains the fix.

Before, with two value types referencing the same buildable type under different attribute names:

| file order | helpers in `ImmutableAlpha` | helpers in `ImmutableBeta` |
|---|---|---|
| `Alpha`, `Beta` | `alphaInner*` | `alphaInner*` |
| `Beta`, `Alpha` | `betaInner*` | `betaInner*` |

After, each type gets its own attribute's name and the output no longer depends on order:

| file order | helpers in `ImmutableAlpha` | helpers in `ImmutableBeta` |
|---|---|---|
| `Alpha`, `Beta` | `alphaInner*` | `betaInner*` |
| `Beta`, `Alpha` | `alphaInner*` | `betaInner*` |

`value-fixture` is unchanged: 325 tests, no failures or errors, matching a baseline run with the fix reverted.

Note that this does not address the other half of the shared-cache behaviour discussed in the issue — a stale negative verdict for the *same* attribute still survives into later rounds, because the map is never cleared. That needs a separate decision about the cache's lifetime, so I left it alone.


pbielicki:
Closing this in favour of #1674, which fixes the same issue differently.

The approach here is wrong. Adding the attribute name to the cache key makes each attribute get its own descriptor, so call sites are emitted per attribute. But the helper definitions come from `ValueType.getUniqueAttributeBuilderAttributes()`, which filters to settable attributes. Any value type with a `@Value.Derived` attribute of a buildable type then calls a helper that is never defined, and the generated code does not compile:

```
EdgeProxyConfigImpl.java:255: error: cannot find symbol
    this.mcp = mcpValueOf(computedValue);
```

That project has three `@Value.Derived` attributes of one buildable type and no settable attribute of it, so no definition is emitted at all. It compiles today only because all attributes of a type currently share one descriptor, so every call site resolves to the single definition emitted for whichever attribute won.

The real defect is narrower than this PR assumes: the helper is emitted once per type but named after an attribute. #1674 names it after the value type instead, using `getQualifiedValueTypeNameWithUnderscores()`, which already existed unreferenced. The cache is left alone, so detection behaviour is unchanged.

PR #1674, opened 2026-07-30
Name attribute builder helpers after the value type

### LINKED ISSUE #1672: Generated code is not reproducible: attribute-builder helper names leak between attributes through a static cache

pbielicki:
Fixes #1672.

Attribute builder helpers are emitted once per buildable type but named after an attribute, and which attribute wins depends on which one the processor analysed first — `AttributeBuilderReflection.analyzedReturnTypes` is a static cache keyed by the contained type and never cleared. Identical sources then generate different code between runs, and `ImmutableBeta` can end up with helpers named after an attribute of `Alpha`.

Naming them after the value type removes the ordering dependency. `AttributeBuilderDescriptor.getQualifiedValueTypeNameWithUnderscores()` already existed for this and was unreferenced.

The uniqueness filter had to move to the same key. It deduplicated on the whole descriptor, which is finer: an attribute declared with the abstract type and one declared with the generated immutable type keep separate descriptors, so both emit a helper under the same name and the generated source stops compiling. That shape compiles on master, so changing the name without changing the key would have broken it on upgrade.

Two tests — a fixture reaching one buildable type under three attribute names, one of them `@Value.Derived`; and a two-stage compilation test for the duplicate helper, because a type is only detected as buildable once its immutable exists as a class file. `value-fixture` passes with 327 tests.

Supersedes #1673, which added the attribute name to the cache key instead. That breaks value types with a `@Value.Derived` attribute of a buildable type: call sites are generated for every attribute-builder attribute, but definitions only for settable ones.

### DIFF
--- a/processor-testlib/test/org/immutables/value/processor/meta/AttributeBuilderHelperNamingTest.java
+++ b/processor-testlib/test/org/immutables/value/processor/meta/AttributeBuilderHelperNamingTest.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2026 Immutables Authors and Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.immutables.value.processor.meta;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import javax.tools.Diagnostic;
+import javax.tools.DiagnosticCollector;
+import javax.tools.JavaCompiler;
+import javax.tools.JavaFileObject;
+import javax.tools.SimpleJavaFileObject;
+import javax.tools.StandardJavaFileManager;
+import javax.tools.StandardLocation;
+import javax.tools.ToolProvider;
+import org.immutables.value.Value;
+import org.immutables.value.processor.Processor;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import static org.immutables.check.Checkers.check;
+
+/**
+ * Attribute builder helpers are emitted once per value type and named after it, so two attributes
+ * reaching one value type by different routes have to resolve to a single helper. Here {@code Foo}
+ * is reached through the abstract type and through the generated {@code ImmutableFoo}, which yields
+ * two descriptors carrying the same value type name. Deduplicating helper definitions on anything
+ * finer than that name emits both under one signature, and the generated source stops compiling.
+ * <p>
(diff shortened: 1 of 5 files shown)


elucash:
Thank you for the PR! fixing those parts are much appreciated (attribute builders haven't received much attention in review/verification). It looks and sounds good, surely hope this in better state now. Thank you!

PR #1679, opened 2026-09-11
Fix #1678: order absent values instead of dereferencing them

### LINKED ISSUE #1678: In-memory backend throws NullPointerException when ordering by a nullable attribute
Affects: 2.12.2 (latest release) and master — criteria-inmemory (InMemoryBackend)

Description:

Ordering a query by an attribute that is absent on at least one stored value fails with a NullPointerException instead of returning an order. InMemoryBackend.Session.query() builds its comparator with Comparator.comparing(fn), which has no null handling, so the first extracted null raises inside TimSort.

This affects every nullable attribute, including nested ones where an intermediate object is absent and ReflectionExtractor returns null for the whole path. It is also reachable through a compound sort whose first key contains no nulls at all: a tie on the first key falls through to a second key that does.

Worth noting the asymmetry with filtering. A @Nullable attribute generates an OptionalStringMatcher, so absence is a first-class, typed question when filtering (isAbsent()) and an unguarded null when sorting. Sorting a nullable column is an ordinary operation in any browsable list, not an edge case.

Steps to reproduce:

// a type with @Nullable String nullable()
repository.insert(generator.get().withId("id1").withNullable("a"));
repository.insert(generator.get().withId("id2").withNullable(null));
repository.insert(generator.get().withId("id3").withNullable("b"));

repository.findAll().orderBy(criteria.nullable.asc()).fetch();

Actual: NullPointerException from java.util.TimSort.

Expected: the values ordered, with missing values first when ascending and last when descending — the order MongoDB produces for the same data, so that a database query and an in-memory evaluation of the same criterion agree.

I will deliver a pull request for this issue.

rgatbert:
Ordering by a nullable attribute raised NullPointerException instead of returning an order: the sort comparator was built with Comparator.comparing and no null handling, so the first absent value failed inside TimSort.

The comparator now compares values through nullsFirst(naturalOrder()). Absent values sort first when ascending and, since the comparator is reversed for a descending key, last when descending — the order MongoDB produces for a null or missing attribute, so a database query and an in-memory evaluation of the same criterion agree.

The first commit is the failing test. It covers ascending, descending, and a compound sort whose first key holds no null at all — the tie on that key falls through to a second key that does, which is how this is reached from a sort key that looks safe.

Fixes #1678


asereda-gs:
Thank you for this PR.

Can I ask you to add a special `OrderByTemplate` so this ordering can be tested across different backends ? I think it is useful outside InMemoryBackend.

For examples look at `criteria/common/test/org/immutables/criteria/typemodel`.

asereda-gs:
I wonder if we should introduce `NullsOrdering` enum (`NULLS_LAST` / `NULLS_FIRST`)  as an argument to `OrderMatcher.asc()` (and `desc()`)?

Something like

```
name.asc(NullsOrdering.NULLS_LAST)
```

rgatbert:
> Thank you for this PR.
> 
> Can I ask you to add a special `OrderByTemplate` so this ordering can be tested across different backends ? I think it is useful outside InMemoryBackend.
> 
> For examples look at `criteria/common/test/org/immutables/criteria/typemodel`.

Done. OrderByTemplate in typemodel covers ascending/descending, a nullable attribute, and a sort whose second key holds the null. Registered for in-memory, mongo and geode — all three agree: missing first ascending, last descending.

Elasticsearch is left unregistered, and running the template against it is what makes your second point concrete, so let me take them together.

rgatbert:
> I wonder if we should introduce `NullsOrdering` enum (`NULLS_LAST` / `NULLS_FIRST`) as an argument to `OrderMatcher.asc()` (and `desc()`)?
> 
> Something like
> 
> ```
> name.asc(NullsOrdering.NULLS_LAST)
> ```

You were right. I read your comment as a nicety at first. Registering the template for elasticsearch is what changed my mind: it fails ascending and passes descending, because elasticsearch defaults missing to _last. That is its documented behaviour, not a defect, and it means the in-memory backend has no correct default at all. It mirrors whichever database it sits beside.

I would still keep the enum out of this PR. Removing the NullPointerException means the comparator has to put an absent value somewhere, so this PR has to pick a default either way; adding the enum later only adds a way to ask for the other one, and undoes nothing here. Filed as #1682 with the details and two open questions.

asereda-gs:
Elastic search does not work? Or you're saying it  has different defaults for nulls first/last?

PR #1681, opened 2026-09-11
Fix #1680: compare values by value, not by representation

### LINKED ISSUE #1680: In-memory backend compares values by representation, so a stored 2.00 does not match a queried 2
Affects: 2.12.2 (latest release) and master — criteria-inmemory (ExpressionInterpreter)

Description:

ExpressionInterpreter.binaryCall evaluates EQUAL, NOT_EQUAL, IN and NOT_IN with Objects.equals, and CONTAINS with Iterables.contains, which uses equals as well.

For BigDecimal, equals is true only when the scale matches too, so a stored 2.00 is not equal to a queried 2. MongoDB ignores scale and selects the value. The two backends therefore return different result sets for the same criterion, which is the failure a two-engine setup exists to avoid: one engine answers a query against stored data, the other answers the same query in memory against a single changed item, and the two disagree.

Scale is a property of the representation, not of the number. The practical case is a boundary value: read back from a previous page as 2.00 and fed into the same criterion, it silently drops the item in memory while the server keeps it.

The asymmetry sits inside one backend. Ordering goes through Comparable, so the very same two values are tied when sorted and unequal when filtered, with nothing in any type signature to hint at it.

Steps to reproduce:

repository.insert(generator.get().withId("id1").withValue(new BigDecimal("2.00")));
repository.insert(generator.get().withId("id2").withValue(new BigDecimal("3")));

repository.find(criteria.value.is(new BigDecimal("2")));                     // empty
repository.find(criteria.value.in(new BigDecimal("2"), new BigDecimal("3.0"))); // empty

// same for an iterable attribute holding 2.00
repository.find(criteria.list.contains(new BigDecimal("2")));               // empty

Actual: empty in all three cases.

Expected: id1, id1+id2, id1 — what MongoDB returns for the same data.

Over a list of strings the two equalities agree, which is why contains in particular looks correct until a numeric element is involved.

I will deliver a pull request for this issue.

rgatbert:
ExpressionInterpreter evaluated EQUAL, NOT_EQUAL, IN and NOT_IN with Objects.equals, and CONTAINS with Iterables.contains, which uses equals as well. For BigDecimal that makes scale significant, so a stored 2.00 did not match a queried 2 while MongoDB selects it — the two backends returned different result sets for the same criterion.

Two Numbers are now compared numerically, and contains looks an element up with the same equality as everywhere else rather than delegating to Iterables.contains.

Anything that is not a pair of Numbers keeps going through Object.equals. NaN still equals NaN, because the numeric path is only taken once equals has already said no. A value with no exact decimal representation — infinity, an unknown Number implementation — falls back to comparing its double value.

The first commit is the failing test. It pins is, in/notIn and contains; over a list of strings the two equalities agree, which is why contains in particular looked correct until a numeric element was involved.

Fixes #1680


criteria/inmemory/src/org/immutables/criteria/inmemory/ExpressionInterpreter.java
```diff
@@ -271,6 +278,59 @@ private static Object binaryCall(Call call, Object left, Object right) {
...
+     */
+    private static boolean valuesEqual(Object left, Object right) {
+      if (left instanceof Number && right instanceof Number && !left.equals(right)) {
+        return numbersEqual((Number) left, (Number) right);
+      }
+
+      return Objects.equals(left, right);
+    }
+
+    private static boolean numbersEqual(Number left, Number right) {
```
asereda-gs: Can you please extract this and other number comparison into separate Numbers class. I think it can be package-private.

Also please add tests

